PR #3, opened 2017-07-17
Added date formatter for clickhouse day partition

ecyshor:


src/main/scala/com/crobox/clickhouse/partitioning/PartitionDateFormatter.scala
```diff
@@ -0,0 +1,16 @@
+package com.crobox.clickhouse.partitioning
+
+import java.util.Date
+
+import org.joda.time.DateTime
+import org.joda.time.format.DateTimeFormat
+
+object PartitionDateFormatter {
```
f1yegor: don't get why it's `Partition`. Because for (monthly) partitions format is YYYYMM

PR #8, opened 2017-11-27
Stash messages when manager is not yet initialized

ecyshor:
In a short period before receiving the first health checks we could get connection requests that would receive errors. 
We now handle this by stashing all the actor messages until the first health check is received.


sjoerdmulder:
LGTM :+1:

PR #9, opened 2017-12-05
Parsing of inserts moved away from the ClickhouseIndexingSubscriber

avalon12:
ClickhouseIndexingSubscriber parsed and unparsed inserts now both possible

This is a temporary migration path solution, until all parsing is done outside of this library.

### DIFF
--- a/src/main/scala/com/crobox/clickhouse/stream/ClickhouseIndexingSubscriber.scala
+++ b/src/main/scala/com/crobox/clickhouse/stream/ClickhouseIndexingSubscriber.scala
@@ -3,11 +3,6 @@ package com.crobox.clickhouse.stream
 import akka.Done
 import akka.actor.{Actor, ActorRef, ActorRefFactory, Cancellable, PoisonPill, Props, Terminated}
 import com.crobox.clickhouse.ClickhouseClient
-import com.crobox.clickhouse.stream.ClickhouseBulkActor.Insert
-import com.fasterxml.jackson.annotation.JsonInclude
-import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.datatype.joda.JodaModule
-import com.fasterxml.jackson.module.scala.DefaultScalaModule
 import com.typesafe.scalalogging.LazyLogging
 import org.reactivestreams.{Subscriber, Subscription}
 
@@ -160,7 +155,7 @@ object ClickhouseBulkActor {
 
   case class Request(size: Int)
 
-  case class Insert(table: String, data: Map[String, String])
+  case class Insert(table: String, jsonRow: String)
 
   case class Send(req: String, attempts: Int)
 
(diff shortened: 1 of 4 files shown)


src/main/scala/com/crobox/clickhouse/stream/ClickhouseIndexingSubscriber.scala
```diff
@@ -279,35 +295,70 @@ class ClickhouseBulkActor(targetTable: String,
...
-
   private val objectMapper = new ObjectMapper()
     .registerModule(new JodaModule)
     .registerModule(new DefaultScalaModule)
     .setSerializationInclusion(JsonInclude.Include.NON_DEFAULT)
 
   private def index(): Unit = {
-    val payload = buffer.map(i => objectMapper.writeValueAsString(i.data)).mkString("\n") + "\n"
-    val count   = buffer.size
+    val colList = buffer
```
avalon12: This is a simple but IMO fine implementation that just takes the cols from the biggest element (to cope with empty entries). Should be working as long as all elements have the same shape. If not, then clickhouse will respond with an exception, and it does a good job at it.

src/main/scala/com/crobox/clickhouse/stream/ClickhouseIndexingSubscriber.scala
```diff
@@ -279,35 +295,70 @@ class ClickhouseBulkActor(targetTable: String,
...
+        s"INSERT INTO $table ($colList) VALUES"
+      }
+    }
     client.execute(insertQuery, payload) onComplete {
-      case Failure(_) if retries > 0 => send(payload, count, retries - 1)
-      case Failure(e)                => self ! ClickhouseBulkActor.FlushFailure(e, count)
-      case Success(resp: String)     => self ! ClickhouseBulkActor.FlushSuccess(resp, count)
+      case Failure(_) if retries > 0 =>
+        send(table, payload, count, retries - 1, rawColList)
+      case Failure(e) =>
```
ecyshor: Isn't the retry logic present in the executor as well? Could we reuse it somehow?
Also I do not think that all inserts are retriable, we could retry network failures or unexpected closed connections but should not retry all failures

avalon12: yeah, altho its not really in the scope of this PR... we can make a task for it internally.
Id rather spend more effort on refactoring the clickhouse client when we get to set up the DSL project separately, for then the SOC becomes more apparent.

avalon12: retry logic is present inside the clickhouse client so it is not necesarry here. I took it out

src/main/scala/com/crobox/clickhouse/stream/ClickhouseIndexingSubscriber.scala
```diff
@@ -279,35 +281,49 @@ class ClickhouseBulkActor(targetTable: String,
...
-  private def send(payload: String, count: Int, retries: Int = 3): Unit =
-    client.execute(insertQuery, payload) onComplete {
-      case Failure(_) if retries > 0 => send(payload, count, retries - 1)
-      case Failure(e)                => self ! ClickhouseBulkActor.FlushFailure(e, count)
-      case Success(resp: String)     => self ! ClickhouseBulkActor.FlushSuccess(resp, count)
-    }
+  private def send(table: String, payload: Seq[Map[String, String]], count: Int): Unit = {
+    if (payload.isEmpty) Unit
+
+    val colList = payload.head.keys.mkString(", ")
```
sjoerdmulder: What about the scenario when you insert multiple rows with different columns. f.e.
```scala
Seq(Insert("test", Map("a" -> "b")), Insert("test", Map("b" -> "c")))
```

PR #13, opened 2018-01-25
Added initial dsl to the client

ecyshor:


dsl/src/main/scala/com.crobox.clickhouse/dsl/Query.scala
```diff
@@ -0,0 +1,32 @@
...
+
+case class Limit(size: Long = 100, offset: Long = 0)
+
+trait OrderingDirection
+
+case object ASC extends OrderingDirection
+
+case object DESC extends OrderingDirection
+
+sealed case class UnderlyingQuery(selectQuery: SelectQuery,
```
sjoerdmulder: Better naming would be QueryBuilder?

dsl/src/main/scala/com.crobox.clickhouse/dsl/TableColumn.scala
```diff
@@ -0,0 +1,191 @@
...
+  require(ClickhouseStatement.isValidIdentifier(name), s"Invalid column name identifier")
+
+  def query(): String = s"$name $clickhouseType$defaultValue"
+}
+
+object TableColumn {
+  type AnyTableColumn = Column
+}
+
+import com.crobox.clickhouse.dsl.TableColumn.AnyTableColumn
```
sjoerdmulder: imports at top?

PR #17, opened 2018-02-19
Feature/dsl functional

avalon12:
-Removed all query specific API, and joined it all in the OperationalQuery trait.
This introduces functional immutable Query object, with all transformations from a single facade to wich all transformations result. 

-Removed some duplicate code.

-Queries are now fully composable in any order. Composition functions have been introduced to join multiple separately built queries into a new one. 

-Changed scala string interpolation into Fastring string interpolation, which is at least 3x faster according to benchmarks

-Made TableFromQuery and JoinQuery models more consistent with the rest of the query model


dsl/src/main/scala/com.crobox.clickhouse/dsl/Query.scala
```diff
@@ -21,12 +24,53 @@ case object ASC extends OrderingDirection
...
+      from.orElse(other.from),
+      where.orElse(other.where),
+      if (groupBy.nonEmpty) groupBy else other.groupBy,
+      having.orElse(other.having),
+      join.orElse(other.join),
+      if (orderBy.nonEmpty) orderBy else other.orderBy,
+      limit.orElse(other.limit)
+    )
+
+  def <+:(other: InternalQuery): InternalQuery = :+>(other)
```
sjoerdmulder: Add some documentation about what those expression means? Now it looks like sad santa...

avalon12: added

PR #20, opened 2018-02-19
Introduces CTable...

avalon12:
... an extension on the Table trait for case classes and case objects, 
automatically defining the column list for its own properties.

This could be the default implementation of table, would it not be so that the Product has a limit of 22 values.
When Scala 2.13 is mainstream, we can introduce the HList version as being the default.

### DIFF
--- a/dsl/src/main/scala/com.crobox.clickhouse/dsl/Query.scala
+++ b/dsl/src/main/scala/com.crobox.clickhouse/dsl/Query.scala
@@ -10,6 +10,19 @@ trait Table {
   val columns: List[NativeColumn[_]]
 }
 
+/**
+ * An extension on Table for case classes, 
+ * that provides an implementation of the column list for its own properties.
+ */
+trait CTable extends Table { self: Product =>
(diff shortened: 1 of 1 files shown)


dsl/src/main/scala/com.crobox.clickhouse/dsl/Query.scala
```diff
@@ -10,6 +10,19 @@ trait Table {
   val columns: List[NativeColumn[_]]
 }
 
+/**
+ * An extension on Table for case classes, 
+ * that provides an implementation of the column list for its own properties.
+ */
+trait CTable extends Table { self: Product =>
```
ecyshor: Should we add a simple unit test for this?

avalon12: I think we should, ill add it when I find some time.

avalon12:
I did some testing with the case class and sadly it doesnt work the way I had envisioned. closing this.

PR #21, opened 2018-02-23
Adding ON CLUSTER and Replicated support

sjoerdmulder:


ecyshor:
It would be nice to add the compatibility version for clickhouse to the wiki

PR #27, opened 2018-04-12
Feature/from final

avalon12:


avalon12:
from what i understand it would be interesting to use it only in the inner join as its quite an intensive operation that you would try to keep to a minimum.
Also, I dont see how you would use final in a wrapping join, since its only the table select that you want tto apply the final on, since it will actually try to optimize the table rows, right? lets investigate some more.

PR #31, opened 2018-04-13
Enhance aggregated functions

ecyshor:
- Fixed bug when using timeseries for month/year/quarter, which were dependent on the timezone id and that was not available. For this we are falling back to finding a timezone id with the same offset. There will be differences between the timezone, but if you provide it with an id then that will be used.

- Added all combinators for aggregated functions and cleaned part of dsl

### DIFF
--- a/dsl/src/main/scala/com.crobox.clickhouse/dsl/AggregationFunctions.scala
+++ b/dsl/src/main/scala/com.crobox.clickhouse/dsl/AggregationFunctions.scala
@@ -38,11 +38,11 @@ object AggregateFunction {
      * Having a column with type array, it aggregates all the results for that column by running the provided aggregation functions for each vertical slice of the array elements.
      * Therefore, for the query result:
      * \array_col|
-     * |[x1, y1, z1]
+     * |[x1, y1, z1, u1]
      * |[x2, y2, z2]
      * |[x3, y3, z3]
      *
-     * if you run sumForEach(array_col) you will get an array result with the following entries: [sum(x1,x3,x3), sum(y1,y2,y3), sum(z1, z2, z3)]
+     * if you run sumForEach(array_col) you will get an array result with the following entries: [sum(x1,x3,x3), sum(y1,y2,y3), sum(z1, z2, z3), sum(u1)]
      *
      * */
     def forEach[V, T <: TableColumn[Seq[V]], Res](
--- a/dsl/src/main/scala/com.crobox.clickhouse/dsl/AggregationFunctions.scala
+++ b/dsl/src/main/scala/com.crobox.clickhouse/dsl/AggregationFunctions.scala
@@ -55,6 +55,12 @@ object AggregateFunction {
     def uniq(tableColumn: AnyTableColumn) =
       Uniq(tableColumn)
 
+    def average[T: Numeric](tableColumn: TableColumn[T]) =
+      Avg(tableColumn)
+
+    def any[T](tableColumn: TableColumn[T]) =
+      AnyResult(tableColumn)
+
     def sum[T: Numeric](tableColumn: TableColumn[T]) =
       Sum(tableColumn)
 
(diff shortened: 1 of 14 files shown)


avalon12:
So a few operations like isEmpty not etc. have been removed. I can't seem to find a replacement for these (quite necesarry?) functions. Can you clarify?

ecyshor:
Yes, those were a mistake in the initial setup as they should have never been part of the functions but we should have used `comparison` (might be a nice idea to rethink the naming for this as well). I've now added the support to use comparison everywhere (an comparison has been changed to a subset of  column), so we can now support a broader range of the clickhouse capabilities

dsl/src/main/scala/com.crobox.clickhouse/dsl/AggregationFunctions.scala
```diff
@@ -0,0 +1,259 @@
...
+}
+
+case class Count(column: Option[AnyTableColumn] = None) extends AggregateFunction[Long](column.getOrElse(EmptyColumn()))
+
+case class Uniq(tableColumn: AnyTableColumn, modifier: UniqModifier = Uniq.Normal)
+    extends AggregateFunction[Long](tableColumn)
+
+object Uniq {
+  sealed trait UniqModifier
+  case object Normal   extends UniqModifier
```
sjoerdmulder: Simple or Normal?

sjoerdmulder: Simple or Normal?

dsl/src/main/scala/com.crobox.clickhouse/dsl/AggregationFunctions.scala
```diff
@@ -0,0 +1,259 @@
...
+
+  case class If[T <: Column, Res](condition: Comparison)      extends Combinator[T, Res]
+  case class CombinatorArray[T <: TableColumn[Seq[V]], V]()   extends Combinator[T, V]
+  case class ArrayForEach[T <: TableColumn[Seq[Seq[V]]], V]() extends Combinator[T, Seq[V]]
+  case class State[T <: Column]()                             extends Combinator[T, StateResult]
+  case class Merge[T <: TableColumn[StateResult], Res]()      extends Combinator[T, Res]
+
+  trait AggregationFunctionsCombinersDsl {
+
+    def aggIf[T <: TableColumn[Res], Res](condition: Comparison)(aggregated: AggregateFunction[Res]) =
```
avalon12: Here as well as on other functions wich can be composed out of parts you might consider a flow to work with that more resembles the original query.

Its getting pretty obscure with the current way of defining uniqIf statements for instance, while it could look like this: 

`uniq(if(notEmpty(session_id), MyAggCol)))`

dsl/src/main/scala/com.crobox.clickhouse/dsl/AggregationFunctions.scala
```diff
@@ -0,0 +1,259 @@
...
+
+object Uniq {
+  sealed trait UniqModifier
+  case object Normal   extends UniqModifier
+  case object Combined extends UniqModifier
+  case object HLL12    extends UniqModifier
+  case object Exact    extends UniqModifier
+
+  trait UniqDsl {
+    def combined(uniq: Uniq) = uniq.copy(modifier = Combined)
```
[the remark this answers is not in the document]
ecyshor: We could provide all the possible combinations as clickhouse for uniq/sum/etc.. but having them apply on top as modifiers makes it more flexible for the library user, as it can create his own composed state of the aggregation and use that simply everywhere

dsl/src/main/scala/com.crobox.clickhouse/dsl/AggregationFunctions.scala
ecyshor: The choice for the state combinators is connected to the fact that `if,state,merge` are global available for all the aggregation functions, and the clickhouse system for it is so flexible that it will be impossible to explicitly declare all the support combinations.
Take for example the function `uniqIfIfIfIfIfArrayState`, this would be a valid aggregation function which would have to have 5 if conditions.

PR #38, opened 2018-05-30
Backward compatibile syntax for Engine DSL

AcidFlow:
Backward compatible DSL syntax for MergeTree engines, supporting the old
partitioning from clickhouse (using the month partition from a date column).


avalon12:
LGTM, this will be a temporary solution anyways.

Issue #42, opened 2018-06-19
Queue is full deadlock because of Healthchecker


sjoerdmulder:
If the queue is full and all hosts are unhealthy then the client can never recover. e.g.

`com.crobox.clickhouse.balancing.discovery.health.HostHealthChecker` throws 
`Host http://192.168.x.x:8123 status is DEAD because of exception`
with stacktrace
```
java.lang.RuntimeException: Queue is full
	at com.crobox.clickhouse.internal.ClickHouseExecutor.$anonfun$singleRequest$1(ClickHouseExecutor.scala:50)
	at scala.concurrent.Future.$anonfun$flatMap$1(Future.scala:303)
	at scala.concurrent.impl.Promise.$anonfun$transformWith$1(Promise.scala:37)
	at scala.concurrent.impl.CallbackRunnable.run_aroundBody0(Promise.scala:60)
	at scala.concurrent.impl.CallbackRunnable$AjcClosure1.run(Promise.scala:1)
	at org.aspectj.runtime.reflect.JoinPointImpl.proceed(JoinPointImpl.java:149)
```

Only way to resolve this seems to be to restart the service currently

PR #43, opened 2018-06-27
Health check improvements

ecyshor:
Using non query health checks to verify the status of the server as to not pollute the query log with `SELECT 1` queries.

Also improved logging for health failures.

### DIFF
--- a/client/src/main/scala/com.crobox.clickhouse/balancing/discovery/ConnectionManagerActor.scala
+++ b/client/src/main/scala/com.crobox.clickhouse/balancing/discovery/ConnectionManagerActor.scala
@@ -66,8 +66,8 @@ class ConnectionManagerActor(healthProvider: (Uri) => Props, config: Config)
         logHostStatus(status)
         hostsStatus.put(host, status)
         status.status match {
-          case Alive => connectionIterator.add(host)
-          case Dead  => connectionIterator.remove(host)
+          case Alive   => connectionIterator.add(host)
+          case Dead(_) => connectionIterator.remove(host)
         }
       } else {
         log.info(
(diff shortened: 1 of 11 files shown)


client/src/main/scala/com.crobox.clickhouse/internal/InternalExecutorActor.scala
```diff
@@ -17,11 +18,17 @@ class InternalExecutorActor(override protected val config: Config)
   override implicit val system: ActorSystem = context.system
 
   override def receive = {
-    case Execute(uri: Uri, query: String) =>
+    case Execute(uri: Uri, Some(query)) =>
       val eventualResponse = executeRequest(Future.successful(uri), query)
-      eventualResponse.map(response => response.split("\n").toSeq) pipeTo sender
+      splitResponse(eventualResponse) pipeTo sender
+    case Execute(uri: Uri, None) =>
```
sjoerdmulder: Instead of using `Execute(uri, None)` maybe better to create a separate case case `Healthcheck(uri: String)`?

client/src/main/scala/com.crobox.clickhouse/internal/InternalExecutorActor.scala
```diff
@@ -17,11 +18,17 @@ class InternalExecutorActor(override protected val config: Config)
...
+      val request = HttpRequest(method = HttpMethods.GET, uri = uri)
+      splitResponse(handleResponse(singleRequest(request), "health check", uri)) pipeTo sender
```
sjoerdmulder: Is the `singleRequest` using a different queue? So issue #42 is also fixed then?

client/src/main/scala/com.crobox.clickhouse/balancing/discovery/ConnectionManagerActor.scala
```diff
@@ -66,8 +66,8 @@ class ConnectionManagerActor(healthProvider: (Uri) => Props, config: Config)
         logHostStatus(status)
         hostsStatus.put(host, status)
         status.status match {
-          case Alive => connectionIterator.add(host)
```
sjoerdmulder: Maybe change this to scala `Success` / `Failure` ?

client/src/main/scala/com.crobox.clickhouse/internal/InternalExecutorActor.scala
ecyshor: This uses an internal queue for health/cluster member scanning.

PR #44, opened 2018-06-27
Serving the clickhouse host from configuration

ecyshor:
while the connection manager is initializing you can select to server the host from the configuration

### DIFF
--- a/client/src/main/scala/com.crobox.clickhouse/balancing/discovery/ConnectionManagerActor.scala
+++ b/client/src/main/scala/com.crobox.clickhouse/balancing/discovery/ConnectionManagerActor.scala
@@ -2,6 +2,7 @@ package com.crobox.clickhouse.balancing.discovery
 
 import akka.actor.{Actor, ActorLogging, ActorRef, Cancellable, PoisonPill, Props, Stash, Status}
 import akka.http.scaladsl.model.Uri
+import com.crobox.clickhouse.balancing.HostBalancer
 import com.crobox.clickhouse.balancing.discovery.health.HostHealthChecker.Status.{Alive, Dead}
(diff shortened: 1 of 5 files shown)


client/src/main/scala/com.crobox.clickhouse/balancing/discovery/ConnectionManagerActor.scala
```diff
@@ -48,8 +52,13 @@ class ConnectionManagerActor(healthProvider: (Uri) => Props, config: Config)
...
     case GetConnection() =>
       if (!initialized) {
-        log.warning("Stashing get connection message until connection message is sent to initialize the manager.")
-        stash()
+        if (fallbackToConfigurationHost) {
+          log.warning("Not yet initialized, returning the config host.")
+          sender ! configHost
+        } else {
+          log.warning("Stashing get connection message until connection message is sent to initialize the manager.")
+          stash()
```
avalon12: Is this favorable? Would it possible cause a burst of useless messages being sent after a cluster takes a while to get up or to respond?

ecyshor: We are stashing just until the first connections response is received. 
I do agree that it might not be exactly helpful in some cases but it would not affect the rest of the requests.
We might want to consider the timeout as well and drop the stashed messages after they expire.

avalon12: Yeah id go for that, would make it more durable. Its exactly the logic you would expect from a driver (in terms of graceful degradation in case of trouble)

ecyshor: This would be an improvement to the existing behavior and we should tackle it in another pull request.

PR #47, opened 2018-07-02
Ensuring that we run just one health check at a time per host

ecyshor:
When healthchecks are timing out there were cases where the internal queues were filling with health queries.
We are now only doing at most one health check at the same time per host.


ecyshor:
Should fix https://github.com/crobox/clickhouse-scala-client/issues/42

PR #48, opened 2018-07-06
Periodically logging unreachable hosts

ecyshor:


client/src/main/scala/com.crobox.clickhouse/balancing/discovery/ConnectionManagerActor.scala
```diff
@@ -93,6 +95,12 @@ class ConnectionManagerActor(healthProvider: (Uri) => Props, config: Config)
         }
         log.info("Connection manager initialized")
       }
+
+    case LogDeadConnections =>
+      val deadHosts = hostsStatus.values.collect {
+        case HostStatus(host, Dead(_)) => host
+      }
+      log.error(s"Hosts ${deadHosts.mkString(" - ")} are still unreachable")
```
sjoerdmulder: Maybe only log when there are actually deadHosts e.g. `nonEmpty`?

PR #49, opened 2018-07-09
Added clickhouse sink based fully on akka streams

ecyshor:


client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -0,0 +1,33 @@
...
+import com.crobox.clickhouse.stream.ClickhouseBulkActor.{ClickhouseIndexingException, Insert}
+import com.typesafe.config.Config
+import com.typesafe.scalalogging.LazyLogging
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
+
+object ClickhouseSink extends LazyLogging {
+
+  def insertSink(config: Config, client: ClickhouseClient)(implicit ec: ExecutionContext): Sink[Insert, Future[Done]] = {
```
sjoerdmulder: Maybe we should make the config configurable so that for different sinks you have different configs possible (1 with 1000 inserts batch other with 10000?

PR #50, opened 2018-07-19
Method to derive a TimeUnit from a Joda Period

avalon12:


ecyshor:
LGTM!

client/src/main/scala/com.crobox.clickhouse/time/TimeUnit.scala
```diff
@@ -11,6 +13,18 @@ sealed trait TimeUnit {
       case unit: MultiTimeUnit =>
         MultiDuration(value, unit)
     }
+
+  def apply(period: Period): Option[TimeUnit] = period.toString match {
+    case "PT1S" => Some(TimeUnit.Second)
```
sjoerdmulder: What if Period is `PT2S`? or this impossible?

avalon12: Then you'd get a None. Such a period can not be expressed as a time unit. Joda library does not really have unit of time like this. Period is the closest thing.

PR #52, opened 2018-07-31
Deterministic intervals for TimeSeries

avalon12:
TimeSeries will now shift back the provided timestamp to the start of the provided interval in a deterministic way. 

This allows for both persisting as well as selecting the same intervals within a multi interval under any condition.

The reference point for all intervals is that of the "GetRelativeXXXNum" functions in clickhouse.

### DIFF
--- a/client/src/main/scala/com.crobox.clickhouse/time/MultiInterval.scala
+++ b/client/src/main/scala/com.crobox.clickhouse/time/MultiInterval.scala
@@ -80,25 +80,9 @@ object MultiInterval {
 
         ref.withMillis((detWeeks * Week.standardMillis) + (Day.standardMillis * 4) - tzOffset)
       case MultiDuration(value, Month) =>
-        val ref = start.withTimeAtStartOfDay.withDayOfMonth(1)
-
-        val months = (ref.getYear * 12) + ref.getMonthOfYear
-        val detRelMonths = (months - 1) - (months % value)
-
-        val detMonthOfYearZeroBased = detRelMonths % 12
-        val detYear = (detRelMonths - detMonthOfYearZeroBased) / 12
-
-        ref.withYear(detYear).withMonthOfYear(detMonthOfYearZeroBased + 1)
+        toStartOfMonth(start, value)
       case MultiDuration(value, Quarter) =>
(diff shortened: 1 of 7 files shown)


dsl/src/main/scala/com.crobox.clickhouse/dsl/AggregationFunctions.scala
```diff
@@ -82,10 +82,20 @@ object AggregateFunction {
...
+      *
+      * Meaning that as long as the duration is the same, your groups will be in the same from/to timestamps
+      *
+      * This is useful for aggregating results by periods of time (group by month, 2 months, days, etc.)
+      *
+      * @param tableColumn
+      * @param interval
+      * @return
+      */
+    def timeSeries(tableColumn: TableColumn[Long], interval: MultiInterval) =
```
ecyshor: Can we add a segment in the readme about time grouping? 
And some examples with code usages and expected multi interval deterministic behavior?

avalon12: I will do that right away (add it to the wiki). If everything else looks ok please merge =x

client/src/main/scala/com.crobox.clickhouse/time/MultiInterval.scala
```diff
@@ -33,38 +33,78 @@ object MultiInterval {
...
+
+        val days = (ref.getMillis + tzOffset) / Day.standardMillis
+        val detDays = days - (days % value)
+
+        ref.withMillis((detDays * Day.standardMillis) - tzOffset)
+      case MultiDuration(value, Week) =>
+        val ref = start.withTimeAtStartOfDay.withDayOfWeek(DateTimeConstants.MONDAY)
+        val tzOffset = ref.getZone.getOffset(ref.withZone(DateTimeZone.UTC))
+
+        val msWeek1 = ref.getMillis - (Day.standardMillis * 4) + tzOffset
```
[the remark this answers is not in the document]
avalon12: the result is stored as msWeek1, wich represents the millis since week 1
I added some comments above this variable to explain whats happening

PR #53, opened 2018-08-15
Added string functions (split and concat array)

ecyshor:
Building insert table in the akka stream sink

### DIFF
--- a/dsl/src/main/scala/com.crobox.clickhouse/dsl/language/ClickhouseTokenizerModule.scala
+++ b/dsl/src/main/scala/com.crobox.clickhouse/dsl/language/ClickhouseTokenizerModule.scala
@@ -100,7 +100,7 @@ trait ClickhouseTokenizerModule extends TokenizerModule {
         if (separator.length == 1) {
           fast"splitByChar(${StringQueryValue(separator)},${tokenizeColumn(tableColumn)})"
         } else {
-          fast"splitByString(${StringQueryValue(separator)} ,${tokenizeColumn(tableColumn)})"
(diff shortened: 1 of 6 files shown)


dsl/src/main/scala/com.crobox.clickhouse/dsl/language/ClickhouseTokenizerModule.scala
```diff
@@ -95,8 +94,16 @@ trait ClickhouseTokenizerModule extends TokenizerModule {
 
   private def tokenizeExpressionColumn(col: ExpressionColumn[_]): String =
     col match {
-      case agg: AggregateFunction[_]    => tokenizeAggregateFunction(agg)
-      case ArrayJoin(tableColumn)       => fast"arrayJoin(${tokenizeColumn(tableColumn)})"
+      case agg: AggregateFunction[_] => tokenizeAggregateFunction(agg)
+      case ArrayJoin(tableColumn)    => fast"arrayJoin(${tokenizeColumn(tableColumn)})"
+      case SplitString(tableColumn, separator) =>
+        if (separator.length == 1) {
+          fast"splitByChar(${StringQueryValue(separator)},${tokenizeColumn(tableColumn)})"
```
sjoerdmulder: Inconsisten spacing, triggering OCD :smiling_imp:

ecyshor: It seems that that's the way scalafmt formats it, we could update the rules to try a fix.

PR #54, opened 2018-08-15
Added type propagation to the aggregate function combinators

avalon12:


dsl/src/main/scala/com.crobox.clickhouse/dsl/AggregationFunctions.scala
```diff
@@ -29,10 +29,10 @@ object AggregateFunction {
 
   trait AggregationFunctionsCombinersDsl {
 
-    def aggIf[T <: TableColumn[Res], Res](condition: Comparison)(aggregated: AggregateFunction[Res]) =
+    def aggIf[T <: TableColumn[Res], Res](condition: Comparison)(aggregated: AggregateFunction[Res]): CombinedAggregatedFunction[T, Res] =
```
ecyshor: Shouldn't we set the return type to the parent `AggregateFunction` instead of `CombinedAggregatedFunction`?  It's not useful to have the combined one returned

avalon12: it has to, because the tokenizer needs to know the combinator to tokenize

PR #55, opened 2018-08-16
temporary DSL for arithmetic functions

avalon12:
until we get the full DSL function set from the vocabulary update.

### DIFF
--- a/dsl/src/main/scala/com.crobox.clickhouse/dsl/TableColumn.scala
+++ b/dsl/src/main/scala/com.crobox.clickhouse/dsl/TableColumn.scala
@@ -127,7 +127,7 @@ case class Const[V: QueryValue](const: V) extends ExpressionColumn[V](EmptyColum
   val parsed = implicitly[QueryValue[V]].apply(const)
 }
 
-trait ColumnOperations extends AggregationFunctionsDsl with TypeCastColumnOperations {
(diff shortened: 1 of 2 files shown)


dsl/src/main/scala/com.crobox.clickhouse/dsl/TableColumn.scala
```diff
@@ -49,6 +49,14 @@ abstract class TypeCastColumn[V](targetColumn: AnyTableColumn) extends Expressio
 case class Reinterpret[V](typeCastColumn: TypeCastColumn[V] with Reinterpretable)
     extends TypeCastColumn[V](typeCastColumn)
 
+
+//FIXME: Here until we add them to the DSL trough the vocabulary update (feature/voc-typed-magnets)
+abstract class ArithmeticFunction[V]() extends ExpressionColumn[V](EmptyColumn())
+case class Multiply[V](left: AnyTableColumn, right: AnyTableColumn) extends ArithmeticFunction[V]
```
ecyshor: Shouldn't left and right be restricted to numeric?

avalon12: you can also work on dates, but we could skip that for now.. for me this is really a quick and dirty setup. 

Want me to change it?

ecyshor: Good for dictionary update

PR #58, opened 2018-09-03
Sending progress from headers in the progress queue

ecyshor:
Added custom akka http client transport which receives all the streaming parts of the http request/response cycle.
Use the client transport to identify the queries by a custom internal header, and intercept the clickhouse progress headers and route them to a global progress queue, from where the events are consumed by individual query progress queue.

Currently we modify the http response and send the end of headers mark as soon as we see the first progress header, but it might be worth it to just identify the response status code and publish the query accepted event directly instead. This way we would reduce the complexity which is introduce by the manipulation of the http response.

### DIFF
--- a/client/src/main/scala/com.crobox.clickhouse/internal/ClickHouseExecutor.scala
+++ b/client/src/main/scala/com.crobox.clickhouse/internal/ClickHouseExecutor.scala
@@ -2,19 +2,23 @@ package com.crobox.clickhouse.internal
 
 import akka.Done
 import akka.actor.ActorSystem
-import akka.http.scaladsl.Http
 import akka.http.scaladsl.model.Uri.Query
 import akka.http.scaladsl.model._
+import akka.http.scaladsl.settings.{ClientConnectionSettings, ConnectionPoolSettings}
+import akka.http.scaladsl.{ClientTransport, Http}
 import akka.stream._
-import akka.stream.scaladsl.{Keep, Sink, Source, SourceQueueWithComplete}
+import akka.stream.scaladsl.{BidiFlow, BroadcastHub, Flow, Keep, Sink, Source, SourceQueue, SourceQueueWithComplete}
+import akka.stream.stage._
+import akka.util.ByteString
 import com.crobox.clickhouse.balancing.HostBalancer
 import com.crobox.clickhouse.internal.ClickHouseExecutor.QuerySettings.ReadOnlySetting
-import com.crobox.clickhouse.internal.ClickHouseExecutor.{QueryProgress, QueryRetry, QuerySettings}
+import com.crobox.clickhouse.internal.ClickHouseExecutor._
 import com.typesafe.config.Config
 import com.typesafe.scalalogging.LazyLogging
 
 import scala.concurrent.{Future, Promise}
-import scala.util.{Failure, Success, Try}
+import scala.util.parsing.json.JSON
+import scala.util.{Failure, Random, Success, Try}
 
 private[clickhouse] trait ClickHouseExecutor extends LazyLogging {
   this: ClickhouseResponseParser with ClickhouseQueryBuilder =>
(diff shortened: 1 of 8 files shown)


client/src/main/scala/com.crobox.clickhouse/internal/ClickHouseExecutor.scala
```diff
@@ -24,9 +28,58 @@ private[clickhouse] trait ClickHouseExecutor extends LazyLogging {
...
+            None
+
+        }
+      })
+      .collect {
+        case Some(progress) => progress
+      }
+      .withAttributes(ActorAttributes.supervisionStrategy({
+        case ex @ _ =>
+          logger.warn("Failed in the progress streaming", ex)
```
avalon12: Wierd english, perhaps use `Detected failure in the query progress stream, resuming operation.`

client/src/main/scala/com.crobox.clickhouse/internal/ClickHouseExecutor.scala
```diff
@@ -41,12 +94,15 @@ private[clickhouse] trait ClickHouseExecutor extends LazyLogging {
   def executeRequest(query: String,
                      settings: QuerySettings,
                      entity: Option[RequestEntity] = None,
-                     progressQueue: Option[SourceQueueWithComplete[QueryProgress]] = None): Future[String] =
+                     progressQueue: Option[SourceQueueWithComplete[QueryProgress]] = None): Future[String] = {
+    val queryIdentifier = Random.alphanumeric.take(20).mkString("")
```
avalon12: Is this the kind of identifier we want? is this proper randomness?

client/src/main/scala/com.crobox.clickhouse/internal/ClickHouseExecutor.scala
```diff
@@ -41,12 +94,15 @@ private[clickhouse] trait ClickHouseExecutor extends LazyLogging {
...
+    logger.info(s"Executing request with identifier $queryIdentifier")
```
avalon12: Feels like this is debug or trace logging

client/src/main/scala/com.crobox.clickhouse/internal/ClickHouseExecutor.scala
```diff
@@ -149,6 +224,132 @@ object ClickHouseExecutor {
...
+            val byteString = grab(in1)
+            if (byteString.containsSlice(ByteString(InternalQueryIdentifier))) {
+              val incomingString  = byteString.utf8String
+              val responseStrings = incomingString.split(Crlf)
+              val queryIdHeader   = responseStrings.find(_.contains(InternalQueryIdentifier))
+              if (queryIdHeader.isEmpty) {
+                log.warning(s"Could not extract the query id from the containing $incomingString")
+              }
+              if (queryIsInProgress) {
+                log.warning("The previous query was not terminated correctly")
```
avalon12: error?

client/src/main/scala/com.crobox.clickhouse/internal/ClickHouseExecutor.scala
```diff
@@ -24,9 +28,58 @@ private[clickhouse] trait ClickHouseExecutor extends LazyLogging {
...
+  lazy val (progressQueue, progressSource) = {
+    val builtSource = Source
+      .queue[String](1000, OverflowStrategy.dropHead)
+      .map(queryAndProgress => {
+        queryAndProgress.split("\n", 2).toList match {
+          case queryId :: progressJson :: Nil =>
+            Try {
+              val parsedJson = JSON.parseFull(progressJson).map(_.asInstanceOf[Map[String, String]])
+              if (parsedJson.isEmpty || parsedJson.get.size != 3) {
+                throw new IllegalAccessException(s"Cannot extract progress from $parsedJson")
```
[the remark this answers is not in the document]
ecyshor: Suppose to be `IllegalArgumentException`, was too quick on the autocomplete. Will fix it

client/src/main/scala/com.crobox.clickhouse/internal/ClickHouseExecutor.scala
```diff
@@ -24,9 +28,58 @@ private[clickhouse] trait ClickHouseExecutor extends LazyLogging {
...
+                val jsonMap = parsedJson.get
+                ClickhouseQueryProgress(
+                  queryId,
+                  Progress(jsonMap("read_rows").toLong, jsonMap("read_bytes").toLong, jsonMap("total_rows").toLong)
+                )
+              }
+            } match {
+              case Success(value) => Some(value)
+              case Failure(exception) =>
+                logger.warn(s"Failed to parse json $progressJson", exception)
```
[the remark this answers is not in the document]
ecyshor: We should restrict from logging to error unless it's a fatal exception. The error level should be treated as `need to act immediately`, and this is not the case.

client/src/main/scala/com.crobox.clickhouse/internal/ClickHouseExecutor.scala
ecyshor: This is just an internal identifier for now so it should be perfectly fine. We might want to reconsider when we'll expose it to the outside

client/src/main/scala/com.crobox.clickhouse/internal/ClickHouseExecutor.scala
ecyshor: Same as before, it's not a terminal error.

avalon12: it could be? errors on the side of the database could lead to inconsistency / dataloss very quickly.

ecyshor: This does not denote an error on the database level, only an error in the way we handle the progress. The query will still be executed completely

PR #68, opened 2018-10-31
Query settings in the api

ecyshor:
Every execute/query method from the client api now accepts a query settings object which can contain any of the settings documented in [https://clickhouse.yandex/docs/en/operations/settings/](clickhouse settings).


dsl/src/main/scala/com.crobox.clickhouse/dsl/execution/ClickhouseQueryExecutor.scala
```diff
@@ -29,7 +31,8 @@ trait ClickhouseQueryExecutor extends QueryExecutor { self: TokenizerModule =>
 
   def executeWithProgress[V: JsonReader](
       query: Query
-  )(implicit executionContext: ExecutionContext): Source[ClickHouseExecutor.QueryProgress, Future[QueryResult[V]]] =
+  )(implicit executionContext: ExecutionContext,
+    settings: QuerySettings = QuerySettings()): Source[QueryProgress, Future[QueryResult[V]]] =
```
sjoerdmulder: Should we provide a default-value for the settings?

ecyshor: The default settings are empty and have a fallback to config settings, so it makes sense to provide this default as in most use cases you do not want to override the config settings

PR #69, opened 2018-10-31
Added distributed engine

ecyshor:


dsl/src/main/scala/com.crobox.clickhouse/dsl/schemabuilder/Engine.scala
```diff
@@ -26,6 +26,15 @@ object Engine {
     override def toString: String = "Memory"
   }
 
+  /**
+   * https://clickhouse.yandex/docs/en/operations/table_engines/distributed/
+   * */
+  case class DistributedEngine(targetTable: String, cluster: String, database: String, shardingKey: Option[String])
```
sjoerdmulder: Keep the same argument order as Clickhouse?

PR #78, opened 2018-11-16
Feature/retry idempotent queries

ecyshor:
- Added the ability to specify that a query is idempotent and safe to be retried (all read queries and for example inserts into replacingmergetree table engines)
- Retrying idempotent queries on all the exceptions
- Using get to execute read queries whenever possible
- When using `sink` we no longer call execute (which applies the retries), but do a single call directly to stream the source as body. The user should handle retries in that case by using akka stream retries.

In the future we might want to make the retry mechanism smarter and add an exponential back off but currently we retry each time on the next host from the balancer so it should be safe.

### DIFF
--- a/client/src/main/scala/com.crobox.clickhouse/internal/ClickhouseQueryBuilder.scala
+++ b/client/src/main/scala/com.crobox.clickhouse/internal/ClickhouseQueryBuilder.scala
@@ -18,6 +18,7 @@ private[clickhouse] trait ClickhouseQueryBuilder extends LazyLogging {
     import akka.http.scaladsl.model.headers.`Accept-Encoding`
     immutable.Seq(`Accept-Encoding`(gzip, deflate))
   }
+  private val MaxUriSize = 16 * 1024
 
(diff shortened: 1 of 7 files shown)


client/src/main/scala/com.crobox.clickhouse/internal/ClickhouseQueryBuilder.scala
```diff
@@ -22,37 +23,44 @@ private[clickhouse] trait ClickhouseQueryBuilder extends LazyLogging {
...
+          uri = urlQuery,
           entity = e,
           headers = Headers ++ queryIdentifier.map(RawHeader(ProgressHeadersAsEventsStage.InternalQueryIdentifier, _))
         )
-      case None if !idempotent =>
+      case None
+          if settings.idempotent && settings.readOnly == ReadQueries && urlQuery
+            .toString()
+            .getBytes
+            .length < 16 * 1024 => //max url size
```
sjoerdmulder: Magic number, make this a config setting?

PR #81, opened 2019-01-03
Adding support for WITH ROLLUP and WITH CUBE

sjoerdmulder:
Also various fixes for validating column names preventing possible SQL injections


dsl/src/main/scala/com.crobox.clickhouse/dsl/TableColumn.scala
```diff
@@ -5,14 +5,14 @@ import com.crobox.clickhouse.dsl.marshalling.QueryValue
 import com.crobox.clickhouse.dsl.schemabuilder.{ColumnType, DefaultValue}
 import com.dongxiguo.fastring.Fastring.Implicits._
 
-sealed case class EmptyColumn() extends TableColumn("")
+case object EmptyColumn extends TableColumn("NULL")
```
ecyshor: Should we rename this to null column in this case?

sjoerdmulder: We could; Ideally we get rid of this thing; it feels a bit like Option, but then misused in certain places

sjoerdmulder: Reason i made it NULL is that empty string is not a valid name for a column; since i added the require(isValidIdentifier) since someone showed me that using ref() you can add `(` and `)` to GROUP BY queries so it already supports `GROUP BY ROLLUP(a,b,c)`

ecyshor:
LGTM

Issue #94, opened 2020-03-18
Scala 2.13 support


zella:
Any plans about it?

sjoerdmulder:
Yes! But for now it has low priority. If you want to create a PR that would great!

sjoerdmulder:
I remember again, i tried it a couple of months back but i got stuck because of a dependency not being available for 2.13... See: https://github.com/Atry/fastring/issues/30

sjoerdmulder:
I removed fastring and tried to compile with 2.13 but there are some deeper issues with the test... if you could pick it up / continue that would be great! :+1:

sjoerdmulder:
Okay! release 0.9.2 is done with support for scala 2.13 :)

PR #95, opened 2020-03-20
WIP: compile with 2.12, 2.13

zella:
#94 Hi, I successfully compile with `implicit ClassTag[T]`.  Here problems, where I can't use `ClassTag[T]`, so Implicits not worked in some cases. Anyway I create pr to share experience and to speedup resolving issue. Help would be nice.

### DIFF
--- a/dsl/src/test/scala/com/crobox/clickhouse/dsl/language/CollectionFunctions.scala
+++ b/dsl/src/test/scala/com/crobox/clickhouse/dsl/language/CollectionFunctions.scala
@@ -22,12 +22,7 @@ class CollectionFunctions extends ColumnFunctionTest {
 
   it should "succeed for IN functions" in {
     val someCollection = Seq(1,4,6,9)
-    val someTuple = tuple(
-      constOrColMagnetFromConst(1),
-      constOrColMagnetFromConst(4),
-      constOrColMagnetFromConst(6),
-      constOrColMagnetFromConst(9)
-    )
+    val someTuple = tuple(1: Int, 4: Int, 6: Int, 9: Int)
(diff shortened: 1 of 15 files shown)


sjoerdmulder:
Thanks! Will take a look at it too!

dsl/src/test/scala/com/crobox/clickhouse/dsl/language/CollectionFunctions.scala
```diff
@@ -22,7 +22,7 @@ class CollectionFunctions extends ColumnFunctionTest {
 
   it should "succeed for IN functions" in {
     val someCollection = Seq(1,4,6,9)
-    val someTuple = tuple(1,4,6,9)
+    val someTuple = tuple(1: Int, 4: Int, 6: Int, 9: Int)
```
sjoerdmulder: I don't think we want to have those type hints, are they required to make it compile?

zella: Yes. Note that only type hints make it compile in 2.13.

sjoerdmulder: Ah, I Found what the issue was that doesn't require any changes to the test. This must be a requirement since you don't want todo that in all you implementation code. Opened PR #97 that adds scala 2.13 support.

sjoerdmulder:
Thanks a lot for your work!  I have cherry picked some of it in a seperate branch because of ease of use :)

0.9.2 is released with 2.13 support

PR #96, opened 2020-03-26
Feature/array functions

lwolters:
- Added support for missing ArrayFunctions: has, hasAll, hasAny and arrayResize
- Added normal tests and integration tests

### DIFF
--- a/dsl/src/main/scala/com.crobox.clickhouse/dsl/column/ArrayFunctions.scala
+++ b/dsl/src/main/scala/com.crobox.clickhouse/dsl/column/ArrayFunctions.scala
@@ -6,80 +6,104 @@ import org.joda.time.{DateTime, LocalDate}
 trait ArrayFunctions { this: Magnets =>
   sealed trait ArrayFunction
 
-  abstract class ArrayFunctionOp[V]
-    extends ExpressionColumn[V](EmptyColumn)
-    with ArrayFunction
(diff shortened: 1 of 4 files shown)


dsl/src/main/scala/com.crobox.clickhouse/dsl/column/ArrayFunctions.scala
```diff
@@ -6,74 +6,104 @@ import org.joda.time.{DateTime, LocalDate}
...
-  case class EmptyArrayInt16()  extends ArrayFunctionConst[Long]
-  case class EmptyArrayInt32()  extends ArrayFunctionConst[Long]
-  case class EmptyArrayInt64()  extends ArrayFunctionConst[Long]
-  case class EmptyArrayFloat32() extends ArrayFunctionConst[Float]
-  case class EmptyArrayFloat64() extends ArrayFunctionConst[Float]
-  case class EmptyArrayDate()   extends ArrayFunctionConst[LocalDate]
-  case class EmptyArrayDateTime() extends ArrayFunctionConst[DateTime]
-  case class EmptyArrayString() extends ArrayFunctionConst[String]
-  case class Range(n: NumericCol[_])   extends ArrayFunctionConst[Long]
+  case class EmptyArrayUInt8()       extends ArrayFunctionConst[Long]
```
sjoerdmulder: Are those types correct? Since Int32 should be `Int` and Float64 is a `Double`

PR #104, opened 2020-08-04
Feature/altinity release

lwolters:
Added full support for ALIAS'es for JOIN

### DIFF
--- a/dsl/src/main/scala/com.crobox.clickhouse/dsl/OperationalQuery.scala
+++ b/dsl/src/main/scala/com.crobox.clickhouse/dsl/OperationalQuery.scala
@@ -42,14 +42,13 @@ trait OperationalQuery extends Query {
     OperationalQuery(internalQuery.copy(from = Some(from)))
   }
 
-  def asFinal: OperationalQuery = {
+  def asFinal: OperationalQuery =
     OperationalQuery(internalQuery.copy(asFinal = true))
-  }
 
   def groupBy(columns: Column*): OperationalQuery = {
     val internalGroupBy = internalQuery.groupBy.getOrElse(GroupByQuery())
-    val newGroupBy = Some(internalGroupBy.copy(usingColumns = internalGroupBy.usingColumns ++ columns))
-    val newSelect = mergeOperationalColumns(columns)
+    val newGroupBy      = Some(internalGroupBy.copy(usingColumns = internalGroupBy.usingColumns ++ columns))
+    val newSelect       = mergeOperationalColumns(columns)
     OperationalQuery(
       internalQuery.copy(select = newSelect, groupBy = newGroupBy)
     )
@@ -96,9 +95,11 @@ trait OperationalQuery extends Query {
   def limit(limit: Option[Limit]): OperationalQuery =
     OperationalQuery(internalQuery.copy(limit = limit))
 
-  def unionAll(otherQuery : OperationalQuery): OperationalQuery = {
-    require(internalQuery.select.isDefined && otherQuery.internalQuery.select.isDefined, "Trying to apply UNION ALL on non SELECT queries.")
-    require(otherQuery.internalQuery.select.get.columns.size == internalQuery.select.get.columns.size,
+  def unionAll(otherQuery: OperationalQuery): OperationalQuery = {
+    require(internalQuery.select.isDefined && otherQuery.internalQuery.select.isDefined,
+            "Trying to apply UNION ALL on non SELECT queries.")
+    require(
+      otherQuery.internalQuery.select.get.columns.size == internalQuery.select.get.columns.size,
       "SELECT queries needs to have the same number of columns to perform UNION ALL."
     )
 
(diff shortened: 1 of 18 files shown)


dsl/src/main/scala/com.crobox.clickhouse/dsl/misc/RandomStringGenerator.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import java.security.SecureRandom
+
+import scala.annotation.tailrec
+
+object RandomStringGenerator {
+  private lazy val numberGenerator: SecureRandom = new SecureRandom
+
+  def random(length: Int = 6): String =
+    padLeft(Integer.toString(numberGenerator.nextInt(Int.MaxValue - 1) + 1, 36), length)
```
sjoerdmulder, line 11: So it's a single random char with 5 '0' by default?

sjoerdmulder: Maybe want to reuse the `Random.alphanumeric.take(length).mkString("")` instead?

dsl/src/test/scala/com/crobox/clickhouse/dsl/JoinQueryIT.scala
```diff
@@ -53,7 +54,7 @@ class JoinQueryIT
   ) { (joinType, result) =>
     ignore should s"join correctly on: $joinType" in {
       val query: OperationalQuery =
-        select(itemId).from(select(itemId).from(TwoTestTable).join(joinType, ThreeTestTable).using(itemId))
+        select(itemId).from(select(itemId).from(TwoTestTable).join(joinType, ThreeTestTable, Option("TTT")).using(itemId))
```
sjoerdmulder, line 57: This test is ignored?

dsl/src/test/scala/com/crobox/clickhouse/dsl/JoinQueryIT.scala
```diff
@@ -38,7 +39,7 @@ class JoinQueryIT
   ) { (joinType, result) =>
     it should s"join correctly on: $joinType" in {
       val query: OperationalQuery =
-        select(itemId).from(select(itemId).from(TwoTestTable).join(joinType, ThreeTestTable).using(itemId))
+        select(itemId).from(select(itemId).from(TwoTestTable).join(joinType, ThreeTestTable, Option("TTT")).using(itemId))
```
sjoerdmulder, line 42: How do we prevent that some creates an invalid query by having 2 join's with the same alias?

dsl/src/main/scala/com.crobox.clickhouse/dsl/OperationalQuery.scala
```diff
@@ -130,51 +133,103 @@ trait OperationalQuery extends Query {
...
 
-  def join[TargetTable <: Table](`type`: JoinQuery.JoinType, query: OperationalQuery): OperationalQuery = {
-    OperationalQuery(internalQuery.copy(join = Some(JoinQuery(`type`, InnerFromQuery(query)))))
-  }
+  def join[TargetTable <: Table](`type`: JoinQuery.JoinType,
+                                 query: OperationalQuery,
+                                 alias: Option[String]): OperationalQuery =
+    OperationalQuery(
+      internalQuery.copy(
+        join = Some(JoinQuery(`type`, InnerFromQuery(query), alias = alias.getOrElse(RandomStringGenerator.random())))
```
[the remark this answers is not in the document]
lwolters: The difficulty is that a single query might have many joins. Since a join might also involve a 'select query', it's hard to come up with a consistent naming convention. That said, it's up to the implementation of the API to provide an alias. If not explicitly set, a random one will be used.

I think we should keep it as is....

dsl/src/main/scala/com.crobox.clickhouse/dsl/misc/RandomStringGenerator.scala
lwolters: This code is taken from CUID ;-). I'll write some tests for it

dsl/src/test/scala/com/crobox/clickhouse/dsl/JoinQueryIT.scala
lwolters: ASOF is a tricky test since it is higly clickhouse specific. I'll look into it

dsl/src/test/scala/com/crobox/clickhouse/dsl/JoinQueryIT.scala
lwolters: We simply can't. Again, it's up to the implementation to provide unique names.... If omitted, random ones will be used which are 'safe'

dsl/src/test/scala/com/crobox/clickhouse/dsl/JoinQueryIT.scala
lwolters: Fixed it. Also added assume in order for tests to succeed on clickhouse 19 versions

PR #106, opened 2020-08-07
Feature/join automatic alias

lwolters:
Added full 'automatic' support for joins in aliases. Added a TokenizerContext thats stores a counter to keep track of number of joins, which is automatically used in the alias names

Also added a matchSQL scala test matcher to ease tests


dsl/src/main/scala/com.crobox.clickhouse/dsl/OperationalQuery.scala
```diff
@@ -34,18 +32,17 @@ trait OperationalQuery extends Query {
...
-  }
+  def from[T <: Table](table: T): OperationalQuery =
+    OperationalQuery(internalQuery.copy(from = Some(TableFromQuery(table))))
+
+  def from(query: OperationalQuery): OperationalQuery =
+    OperationalQuery(internalQuery.copy(from = Some(InnerFromQuery(query))))
 
   def asFinal: OperationalQuery =
-    OperationalQuery(internalQuery.copy(asFinal = true))
+    OperationalQuery(internalQuery.copy(as = Option("FINAL")))
```
sjoerdmulder, line 45: FINAL is not the same as AS (aliasing) right? How do i now do `SELECT * FROM table FINAL AS other`?

dsl/src/main/scala/com.crobox.clickhouse/dsl/OperationalQuery.scala
```diff
@@ -133,178 +130,86 @@ trait OperationalQuery extends Query {
...
+  }
 
+  def on(columns: Column*): OperationalQuery = {
+    require(internalQuery.join.isDefined)
+    OperationalQuery(
+      internalQuery.copy(join = Some(this.internalQuery.join.get.copy(on = columns.map(c => (c, "=", c)))))
+    )
+  }
+
+  def on(condition: (Column, String, Column), conditions: (Column, String, Column)*): OperationalQuery = {
```
sjoerdmulder, line 210: What is this String value? Shouldn't it be a sealed trait?

dsl/src/main/scala/com.crobox.clickhouse/dsl/language/ClickhouseTokenizerModule.scala
```diff
@@ -97,9 +102,14 @@ trait ClickhouseTokenizerModule
     }
   }
 
-  private def tokenizeFinal(asFinal: Boolean): String = if (asFinal) "FINAL" else ""
+  private def tokenizeAs(as: Option[String]): String =
+    as.map {
+        case "FINAL" => "FINAL"
```
sjoerdmulder, line 107: Hmmm this doesn't look right, FINAL is different

dsl/src/main/scala/com.crobox.clickhouse/dsl/language/ClickhouseTokenizerModule.scala
```diff
@@ -97,9 +102,14 @@ trait ClickhouseTokenizerModule
...
+        case other   => s"AS $other"
```
sjoerdmulder, line 108: I think it's important to escape this string, so it's SQL safe

dsl/src/main/scala/com.crobox.clickhouse/dsl/language/ClickhouseTokenizerModule.scala
```diff
@@ -214,70 +226,79 @@ trait ClickhouseTokenizerModule
...
+        if (using.nonEmpty) {
+          // TOKENIZE USING
+          if (using.size == 1) s"USING ${using.head.name}"
+          else s"USING (${using.map(_.name).mkString(",")})"
+        } else if (query.on.nonEmpty) {
+          // TOKENIZE ON. If the fromClause is a TABLE, we need to check on aliases!
+          "ON " + query.on
+            .map(cond => {
+              val left = verifyOnCondition(select, from, cond._1)
+              s"${ctx.leftAlias}.$left ${cond._2} ${ctx.rightAlias}.${cond._3.name}"
```
sjoerdmulder, line 278: cond should be a case class instead of a Tuple3; now it's not really clear what's what.

dsl/src/test/scala/com/crobox/clickhouse/dsl/JoinQueryIT.scala
```diff
@@ -82,73 +105,63 @@ class JoinQueryIT
...
-            .using(itemId)
-        )
-      val resultRows = chExecutor.execute[Result](query).futureValue.rows
+
+      var query: OperationalQuery =
+        select(itemId, col2)
+          .from(TwoTestTable)
+          .where(notEmpty(itemId))
+          .join(joinType, ThreeTestTable)
+          .on((itemId, "=", itemId), (col2, "<=", col2))
```
sjoerdmulder, line 114: maybe a DSL like `itemId.joinOn(itemId), col2.joinOnSmaller(col2)` would more clear? Also it prevents the `String` items that you want to prevent to keep it SQL safe.

I'm not completly sure what the naming should be though

dsl/src/main/scala/com.crobox.clickhouse/dsl/FromQuery.scala
```diff
@@ -2,8 +2,14 @@ package com.crobox.clickhouse.dsl
 
 sealed trait FromQuery extends Query with OperationalQuery {
   override val internalQuery = InternalQuery(from = Some(this))
+  val alias: Option[String]
+  val `final`:Boolean
```
sjoerdmulder, line 6: meh i really don't like the backticked variables, i know why it's there. But maybe we can name it  `useFinal`?

dsl/src/main/scala/com.crobox.clickhouse/dsl/JoinQuery.scala
```diff
@@ -57,7 +57,19 @@ object JoinQuery {
...
                      other: FromQuery,
-                     //on: Seq[ExpressionColumn[_]] = Seq.empty,
-                     on: Seq[(Column, String, Column)] = Seq.empty,
+                     on: Seq[JoinCondition] = Seq.empty,
                      using: Seq[Column] = Seq.empty,
                      global: Boolean = false)
+
+case class JoinCondition(left: Column, operator: String, right: Column) {
+
+  assert(JoinCondition.SupportedOperators.contains(operator),
```
sjoerdmulder, line 66: `assert` or `require`?

sjoerdmulder: From StackOverflow:  https://stackoverflow.com/questions/24863185/what-is-an-assertionerror-in-which-case-should-i-throw-it-from-my-own-code
 
The meaning of an AssertionError is that something happened that the developer thought was impossible to happen.

dsl/src/main/scala/com.crobox.clickhouse/dsl/OperationalQuery.scala
```diff
@@ -33,16 +33,38 @@ trait OperationalQuery extends Query {
...
+  def from(query: OperationalQuery, alias: String): OperationalQuery =
+    OperationalQuery(internalQuery.copy(from = Some(InnerFromQuery(query, alias = Option(alias)))))
+
   def asFinal: OperationalQuery =
-    OperationalQuery(internalQuery.copy(as = Option("FINAL")))
+    OperationalQuery(
+      internalQuery.from
+        .map {
+          case _: InnerFromQuery =>
+            throw new AssertionError("It's ILLEGAL to set FINAL on a (sub)query FROM query")
```
sjoerdmulder, line 62: IllegalArgumentException? See also comment above

dsl/src/main/scala/com.crobox.clickhouse/dsl/language/ClickhouseTokenizerModule.scala
```diff
@@ -64,19 +64,21 @@ trait ClickhouseTokenizerModule
...
            | ${tokenizeFiltering(where, "WHERE")}
            | ${tokenizeGroupBy(groupBy)}
            | ${tokenizeFiltering(having, "HAVING")}
            | ${tokenizeOrderBy(orderBy)}
            | ${tokenizeLimit(limit)}
-           | ${tokenizeUnionAll(union)}""".trim.stripMargin.replaceAll("\n", "").replaceAll("\r", "")
+           | ${tokenizeUnionAll(union)}""".trim.stripMargin
+          .replaceAll("\n", "")
+          .replaceAll("\r", "")
+          .replaceAll(" +", " ") // replace double (or more) subsequent spaces by one
```
sjoerdmulder, line 81: What if i have a query that looks for `description = ' a  b '`? (ironcly it collapses the spaces in github editor) I think we should not replace any space, \n or \r in this phase because it might affect user input

dsl/src/main/scala/com.crobox.clickhouse/dsl/language/ClickhouseTokenizerModule.scala
lwolters: Added JoinCondition case class

dsl/src/test/scala/com/crobox/clickhouse/dsl/JoinQueryIT.scala
lwolters: Since 'AS OF' joins won't be used that often (very, very custom clickhouse) the default behavior is to join on simple columns, which is supported

dsl/src/main/scala/com.crobox.clickhouse/dsl/language/ClickhouseTokenizerModule.scala
lwolters: Fixed escaping (but removed the tokenizeAs as a whole)

dsl/src/main/scala/com.crobox.clickhouse/dsl/FromQuery.scala
lwolters: Ok Ok ok, changed it to 'finalized' (without quotes)

PR #108, opened 2020-09-17
Feature/UUID not empty

lwolters:
Added rewriting of notEmpty for UUID based columns, since the default behavior is not supported. Right now, notEmpty is rewritten to ' UUID != 0' which works correctly and as expected


dsl/src/main/scala/com.crobox.clickhouse/dsl/language/UUIDFunctionTokenizer.scala
```diff
@@ -0,0 +1,15 @@
+package com.crobox.clickhouse.dsl.language
+
+import com.crobox.clickhouse.dsl._
+
+trait UUIDFunctionTokenizer {
+  self: ClickhouseTokenizerModule =>
+
+  protected def tokenizeUUIDCol(col: UUIDFunction[_])(implicit ctx: TokenizeContext): String = {
```
sjoerdmulder, line 8: This doesn't look right

PR #117, opened 2021-01-08
Feature/clickhouse 20.11.4.13

lwolters:
Updated to latest clickhouse

### DIFF
--- a/dsl/src/main/scala/com.crobox.clickhouse/dsl/column/HigherOrderFunctions.scala
+++ b/dsl/src/main/scala/com.crobox.clickhouse/dsl/column/HigherOrderFunctions.scala
@@ -48,10 +48,10 @@ trait HigherOrderFunctions { self: Magnets =>
                      array: ArrayColMagnet[_ <: Iterable[I]]): ArrayMap[I, O] =
     ArrayMap(Option(func), None, array)
 
-  def arrayMap[I, O](func: (TableColumn[I], TableColumn[I]) => ExpressionColumn[O],
-                     array1: ArrayColMagnet[_ <: Iterable[I]],
-                     array2: ArrayColMagnet[_ <: Iterable[I]]): ArrayMap[I, O] =
-    ArrayMap(None, Option(func), array1, array2)
+//  def arrayMap[I, O](func: (TableColumn[I], TableColumn[I]) => ExpressionColumn[O],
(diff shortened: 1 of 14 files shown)


dsl/src/main/scala/com.crobox.clickhouse/dsl/column/HigherOrderFunctions.scala
```diff
@@ -3,32 +3,240 @@ package com.crobox.clickhouse.dsl.column
...
+                   array2: ArrayColMagnet[_ <: Iterable[I]]): ArrayAll[I, Boolean] =
+    ArrayAll(None, Option(func), None, array1, array2)
+
+  def arrayAll3[I](func: (TableColumn[I], TableColumn[I], TableColumn[I]) => ExpressionColumn[Boolean],
+                   array1: ArrayColMagnet[_ <: Iterable[I]],
+                   array2: ArrayColMagnet[_ <: Iterable[I]],
+                   array3: ArrayColMagnet[_ <: Iterable[I]]): ArrayAll[I, Boolean] =
+    ArrayAll(None, None, Option(func), array1, array2, array3)
+
+  def arrayCount[I](func: Option[TableColumn[I] => ExpressionColumn[Boolean]],
```
sjoerdmulder, line 93: Should we allow this to be `func: Option` since f.e. `arrayAll` and `arrayExists` have no such thing

dsl/src/main/scala/com.crobox.clickhouse/dsl/misc/LogicalOperatorImprovements.scala
```diff
@@ -4,64 +4,64 @@ import com.crobox.clickhouse.dsl._
 
 object LogicalOperatorImprovements {
 
-  implicit class TableColumnImpr(left: TableColumn[Boolean]) {
+  implicit class TableColumnImpr(left: ExpressionColumn[Boolean]) {
```
sjoerdmulder, line 7: For consistency make sure the class is also `ExpressionColumnImpr`; 
But why is this made more strict? e.g. what's wrong with keeping it a `TableColumn`?

lwolters: Reverted to old situation

PR #120, opened 2021-02-16
Feature/multif

lwolters:
Added support for multiIf (and some small improvements)

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "0.10.2"
+version in ThisBuild := "0.10.3-SNAPSHOT"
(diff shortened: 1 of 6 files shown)


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.10.2-SNAPSHOT"
+version in ThisBuild := "0.10.2"
```
sjoerdmulder, line 1: Not sure if we want to remove the -SNAPSHOT

lwolters: no, that shouldn't be the case. Looks like a FU from last deploy.... Good catch

PR #141, opened 2022-11-04
Extend the ClickhouseSink to have support for both Insert and Optimiz…

Majjoodaty:
…e operation

### DIFF
--- a/client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
+++ b/client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
@@ -59,24 +59,28 @@ object ClickhouseSink extends LazyLogging {
           case op: Insert   => Option(op.jsonRow)
           case op: Optimize => optimize = Option(op); None
         }
-        logger.debug(s"Inserting ${payload.size} entries in table: $table.")
-        client
-          .execute(s"INSERT INTO $table FORMAT JSONEachRow", payload.mkString("\n"))
-          .recover {
-            case ex => throw ClickhouseIndexingException("failed to index", ex, payload, table)
-          }
-          .flatMap(result => {
-            optimize
-              .map(o => {
-                val table = o.distributedTable.getOrElse(o.table)
-                client
-                  .execute(s"OPTIMIZE TABLE $table${o.cluster.map(s => s" ON CLUSTER $s").getOrElse("")} FINAL")
-                  .recover {
-                    case ex => throw ClickhouseIndexingException(s"failed to optimize $table", ex, Seq(), table)
-                  }
-              })
-              .getOrElse(Future.successful(result))
+
+        val insertFuture = if (payload.nonEmpty) {
+          logger.debug(s"Inserting ${payload.size} entries in table: $table.")
+          client
+            .execute(s"INSERT INTO $table FORMAT JSONEachRow", payload.mkString("\n"))
+            .recover {
+              case ex => throw ClickhouseIndexingException("failed to index", ex, payload, table)
+            }
+        } else Future.successful("")
+
+        optimize
+          .map(o => {
+            insertFuture.flatMap(_ => {
+              val table = o.distributedTable.getOrElse(o.table)
+              client
+                .execute(s"OPTIMIZE TABLE $table${o.cluster.map(s => s" ON CLUSTER $s").getOrElse("")} FINAL")
+                .recover {
+                  case ex => throw ClickhouseIndexingException(s"failed to optimize $table", ex, Seq(), table)
+                }
+            })
           })
+          .getOrElse(insertFuture)
       })
       .mergeSubstreams
       .toMat(Sink.ignore)(Keep.right)
--- a/client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
+++ b/client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
@@ -59,7 +59,7 @@ object ClickhouseSink extends LazyLogging {
           case op: Insert   => Option(op.jsonRow)
           case op: Optimize => optimize = Option(op); None
         }
-
+        
         if (payload.nonEmpty) {
           optimize match {
             case Some(statement) => insertTable(client, table, payload).flatMap(_ => optimizeTable(client, statement))
@@ -81,17 +81,20 @@ object ClickhouseSink extends LazyLogging {
   private def insertTable(client: ClickhouseClient, table: String, payload: Seq[String])(
       implicit ec: ExecutionContext,
       settings: QuerySettings
-  ): Future[String] = {
-    logger.debug(s"Inserting ${payload.size} entries in table: $table.")
-    client
-      .execute(s"INSERT INTO $table FORMAT JSONEachRow", payload.mkString("\n"))
-      .recover {
-        case ex => throw ClickhouseIndexingException("failed to index", ex, payload, table)
-      }
-  }
+  ): Future[String] =
+    if (payload.nonEmpty) {
+      logger.debug(s"Inserting ${payload.size} entries in table: $table.")
+      client
+        .execute(s"INSERT INTO $table FORMAT JSONEachRow", payload.mkString("\n"))
+        .recover {
+          case ex => throw ClickhouseIndexingException("failed to index", ex, payload, table)
+        }
+    } else Future.successful("")
 
-  private def optimizeTable(client: ClickhouseClient, statement: Optimize)(implicit ec: ExecutionContext,
-                                                                           settings: QuerySettings): Future[String] = {
+  private def optimizeTable(
+      client: ClickhouseClient,
+      statement: Optimize
+  )(implicit ec: ExecutionContext, settings: QuerySettings): Future[String] = {
     val table = statement.distributedTable.getOrElse(statement.table)
     logger.debug(s"Optimizing table: $table.")
     client
--- a/client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
+++ b/client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
@@ -19,7 +19,12 @@ sealed trait TableOperation {
 
 case class Insert(table: String, jsonRow: String) extends TableOperation
 
-case class Optimize(table: String, localTable: Option[String] = None, cluster: Option[String] = None)
+case class Optimize(table: String,
+                    localTable: Option[String] = None,
+                    cluster: Option[String] = None,
+                    partition: Option[String] = None,
+                    `final`: Boolean = true,
+                    deduplicate: Option[String] = None)
     extends TableOperation
 
 object ClickhouseSink extends LazyLogging {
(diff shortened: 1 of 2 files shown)


client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -11,15 +11,30 @@ import scala.concurrent.duration._
...
+sealed trait TableOperation {
+  def table: String
+
+  def jsonRow: String
+
+  def cluster: Option[String]
+}
+
+case class Insert(table: String, jsonRow: String) extends TableOperation {
+  override def cluster: Option[String] = None
```
lwolters, line 25: Removed this cluster statement. It's not used here

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -11,15 +11,30 @@ import scala.concurrent.duration._
...
+  override def jsonRow: String = ""
+}
 
 object ClickhouseSink extends LazyLogging {
 
-  def insertSink(config: Config, client: ClickhouseClient, indexerName: Option[String] = None)(
-      implicit ec: ExecutionContext,
-      settings: QuerySettings = QuerySettings()
-  ): Sink[Insert, Future[Done]] = {
+  def operationSink(config: Config, client: ClickhouseClient, indexerName: Option[String] = None)(
```
lwolters, line 34: In order to maintain backwards compatibility, please still maintain the 'insertSink' method but mark it as deprecated.
In the deprecation warning, please clearly state to use the 'operationSink', which should be renamed to 'toSink'

lwolters, line 34: Rename operationSink to 'toSink'

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -29,24 +44,55 @@ object ClickhouseSink extends LazyLogging {
...
-        logger.debug(s"Inserting ${inserts.size} entries in table: $table. Group Within: ($batchSize - $flushInterval)")
-        client
-          .execute(insertQuery, payload.mkString("\n"))
-          .recover {
-            case ex => throw ClickhouseIndexingException("failed to index", ex, payload, table)
-          }
-          .map(_ => inserts)
+      .mapAsync(mergedIndexerConfig.getInt("concurrent-requests"))(operations => {
+        val table = operations.head.table
+        var insertPayload: Seq[String] = Seq.empty
```
lwolters, line 54: PErhaps make it a mutable.ArrayBuffer?

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -29,24 +44,55 @@ object ClickhouseSink extends LazyLogging {
...
+        var optimizeQuery: Option[String] = None
```
lwolters, line 55: What happens if the batch contains multiple optimize statements (for different tables / cluster)

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -11,15 +11,30 @@ import scala.concurrent.duration._
...
-    extends RuntimeException(msg, cause)
-case class Insert(table: String, jsonRow: String)
+  extends RuntimeException(msg, cause)
+
+sealed trait TableOperation {
+  def table: String
+
+  def jsonRow: String
+
+  def cluster: Option[String]
```
lwolters, line 21: Cluster should be removed from the interface. It should solely be put in the OptimizeOperation since it's only used there

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -29,24 +44,55 @@ object ClickhouseSink extends LazyLogging {
...
+            case Some(o) => optimizeOp(table, o, client)
+            case _ => Future.successful("")
+          })
+          .map(_ => operations)
       })
       .mergeSubstreams
       .toMat(Sink.ignore)(Keep.right)
   }
+
+  private def insertOp(table: String, payload: Seq[String], client: ClickhouseClient)(
```
lwolters, line 74: Please use a sequence of [InsertOperation' instead of json over here. Please maintain logic how to handle the different implementation (of the operation) in place

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -29,24 +44,55 @@ object ClickhouseSink extends LazyLogging {
...
+  ): Future[String] = {
+    val insertQuery = s"INSERT INTO $table FORMAT JSONEachRow"
+    client
+      .execute(insertQuery, payload.mkString("\n"))
+      .recover {
+        case ex => throw ClickhouseIndexingException("failed to index", ex, payload, table)
+      }
+  }
+
+  private def optimizeOp(table: String, query: String, client: ClickhouseClient)(
```
lwolters, line 86: change arguments to a Seq[OptimizeOperation]

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -29,24 +44,55 @@ object ClickhouseSink extends LazyLogging {
...
-          .recover {
-            case ex => throw ClickhouseIndexingException("failed to index", ex, payload, table)
-          }
-          .map(_ => inserts)
+      .mapAsync(mergedIndexerConfig.getInt("concurrent-requests"))(operations => {
+        val table = operations.head.table
+        var insertPayload: Seq[String] = Seq.empty
+        var optimizeQuery: Option[String] = None
+
+        operations.foreach {
```
lwolters, line 57: This setup is sub optimal. Please use a .collect or other way to 'split' the stream/batch of operation into substreams belonging to each type. Fine to use a pattern match 
```
.map {
  case op:InsertOperation => ..
case op:OptimizeOperation => ...
```

Please find the best / most optimized solution

lwolters (changes requested):

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -12,10 +12,22 @@ import scala.concurrent.{ExecutionContext, Future}
...
+
+case class Optimize(table: String, cluster: Option[String]) extends TableOperation
 
 object ClickhouseSink extends LazyLogging {
 
+  /**
+   * @deprecated added support to optimize table operations <br/>
+   *             use [[#toSink()]] instead.
+   */
+  @deprecated
```
lwolters, line 30: @deprecated("use [[#toSink()]] instead")

lwolters: Also,  I was referring to:

def insertSink(config: Config, client: ClickhouseClient, indexerName: Option[String] = None)(
      implicit ec: ExecutionContext,
      settings: QuerySettings = QuerySettings()
  ): Sink[Insert, Future[Done]] = toSink(config, client, indexerName)

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -49,4 +61,69 @@ object ClickhouseSink extends LazyLogging {
...
+        .recover {
+          case ex => throw ClickhouseIndexingException("failed to index", ex, payload, t._1)
+        }
+    })
+
+  private def optimizeOp(ops: Seq[Optimize], client: ClickhouseClient)(
+      implicit ec: ExecutionContext,
+      settings: QuerySettings = QuerySettings()
+  ): Future[Iterable[String]] = {
+    logger.info(s"Optimizing tables")
```
lwolters, line 116: .info? Are you sure? This will generate a LOT of logging

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -49,4 +61,69 @@ object ClickhouseSink extends LazyLogging {
...
+  }
+
+  private def insertOp(ops: Seq[Insert], client: ClickhouseClient)(
+      implicit ec: ExecutionContext,
+      settings: QuerySettings = QuerySettings()
+  ): Future[Iterable[String]] =
+    Future.sequence(ops.groupBy(_.table).map { t =>
+      val insertQuery = s"INSERT INTO ${t._1} FORMAT JSONEachRow"
+      val payload     = t._2.map(_.jsonRow)
+      logger.info(s"Inserting ${payload.size} entries in table: ${t._1}.")
```
lwolters, line 104: .info? Are you sure? This will generate a LOT of logging

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -49,4 +61,69 @@ object ClickhouseSink extends LazyLogging {
...
+    })
+
+  private def optimizeOp(ops: Seq[Optimize], client: ClickhouseClient)(
+      implicit ec: ExecutionContext,
+      settings: QuerySettings = QuerySettings()
+  ): Future[Iterable[String]] = {
+    logger.info(s"Optimizing tables")
+    Future.sequence(
+      ops.map { o =>
+        val query = buildOptimizeQuery(o.table, o.cluster)
```
lwolters, line 119: No need to create a seperate def for a single statement. So directly put the  s"OPTIMIZE TABLE $table${cluster.map(s => s"_local ON CLUSTER $s").getOrElse("")} FINAL" here instead of the function call

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -49,4 +61,69 @@ object ClickhouseSink extends LazyLogging {
...
+      .mergeSubstreams
+      .toMat(Sink.ignore)(Keep.right)
+  }
+
+  private def insertOp(ops: Seq[Insert], client: ClickhouseClient)(
+      implicit ec: ExecutionContext,
+      settings: QuerySettings = QuerySettings()
+  ): Future[Iterable[String]] =
+    Future.sequence(ops.groupBy(_.table).map { t =>
+      val insertQuery = s"INSERT INTO ${t._1} FORMAT JSONEachRow"
```
lwolters, line 102: Try to keep condense! (but still readable of course!)

No need to create a new val insertQuery. Directly put the declartion at line 106.

lwolters (changes requested):

lwolters (approved):

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
```diff
@@ -31,22 +44,50 @@ object ClickhouseSink extends LazyLogging {
...
-            case ex => throw ClickhouseIndexingException("failed to index", ex, payload, table)
-          }
-          .map(_ => inserts)
+      .mapAsync(mergedIndexerConfig.getInt("concurrent-requests"))(operations => {
+        val table = operations.head.table
+        logger.debug(
+          s"Executing ${operations.size} operations on table: $table. Group Within: ($batchSize - $flushInterval)"
+        )
+
+        insertOp(operations.collect { case o: Insert                     => o }, client)
```
lwolters, line 56: Currently, you're traversing the same collection/batch of operations *twice*. This is costly and should be rewritten.
Please google for solutions, if not found, rely on two mutable array buffers (created outside the method and which are cleared in each batch). The big advancement of having these buffers only created once and cleared in each batch is memory consumption

lwolters (changes requested):

client/src/main/scala/com.crobox.clickhouse/stream/ClickhouseSink.scala
Majjoodaty: I changed the code to use ArrayBuffer as suggested. Another solution that can be also used is using the foldLeft to split the batch into two lists like, it will be something like this:

```
        val splitOps = operations.foldLeft((List[Insert](), List[Optimize]()))(
          (x, y) =>
            y match {
              case o: Insert   => (x._1 ++ Seq(o), x._2)
              case o: Optimize => (x._1, x._2 ++ Seq(o))
              case _           => (x._1, x._2)
          }
        )

        insertOp(splitOps._1, client)
          .flatMap(_ => optimizeOp(splitOps._2, client))
          .map(_ => operations)
```

lwolters: Yes, this is indeed a nice solution!

lwolters (approved):

PR #143, opened 2023-04-27
Added support for request compression

elbvs:
For cases when a lot of large insert requests are sent to the server, support for compressing the request to the server itself will be useful

### DIFF
--- a/client/src/test/scala/com/crobox/clickhouse/ClickhouseClientAsyncSpec.scala
+++ b/client/src/test/scala/com/crobox/clickhouse/ClickhouseClientAsyncSpec.scala
@@ -12,7 +12,8 @@ import com.typesafe.config.{Config, ConfigFactory}
 import org.scalatest._
 import org.scalatest.flatspec.AsyncFlatSpecLike
 import org.scalatest.matchers.should.Matchers
-
+import java.io.ByteArrayOutputStream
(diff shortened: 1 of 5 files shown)


client/src/test/scala/com/crobox/clickhouse/ClickhouseClientAsyncSpec.scala
```diff
@@ -58,4 +59,12 @@ abstract class ClickhouseClientAsyncSpec(val config: Config = ConfigFactory.load
         succeed
       })
   }
+
+  def compressGzip(content: String): Array[Byte] = {
+    val arrOutputStream = new ByteArrayOutputStream()
+    val zipOutputStream = new GZIPOutputStream(arrOutputStream)
+    zipOutputStream.write(content.getBytes)
```
sjoerdmulder, line 66: Can we make sure this is using the `UTF-8` encoding? I think by default it will use the system one which might be causing encoding issues

elbvs: Corrected, thanks

sjoerdmulder (approved):

lwolters:
I've released version 1.0.17 including this PR

PR #152, opened 2023-10-31
Added support for first_ and lastValue

lwolters:
Added support for firstValue & lastValue, which are both aggregation functions


lwolters:
Sjoerd did a review and approved it!

PR #158, opened 2024-03-12
Scala 3 migration

aShevc:
This migration is verified by running tests coverage.
Changes in this PR:
- implicit conversions got reworked according to official migration guide
- `Cast` case class and `ComparableWith` trait needed to be moved from mixin traits to Magnet trait, otherwise Scala 3 compiler crash occurred
- as a driveby updated Pekko version

EDIT:
Additional changes in tests:
Due to errors that specifically occured in the CI/CD build, the following changes has been added to tests
- Usages of `===` method in dsl tests replaced by `isEq`, as `===` was evaluated to Boolean
- Moved Type cast method to `scala-2` folder  as it relies on extension methods
- In tests using `InOps` methods implicit conversions were not applied properly. There was a need to explicitly specify type annotation for the fields on which `in` methods were applied. This may require more deep work on this in the future to figure out the root cause of the issue


build.sbt
```diff
@@ -17,7 +17,7 @@ lazy val root = (project in file("."))
       List(
         organization := "com.crobox.clickhouse",
         scalaVersion := "2.13.8",
-        crossScalaVersions := List("2.13.8"),
+        crossScalaVersions := List("2.13.8", "3.3.1"),
```
sjoerdmulder, line 20: Maybe update to latest scala 2.13 also?

dsl/src/main/scala/com/crobox/clickhouse/dsl/misc/DSLImprovements.scala
```diff
@@ -128,7 +128,7 @@ object DSLImprovements {
         //            case _                    =>
         //          }
         case _: TableFromQuery[_] => return query.andConstraint(condition)
-        case _ =>
+        case null =>
```
sjoerdmulder, line 131: Why is this now matching on `null` instead of `_`?

build.sbt
aShevc: sure, will do

dsl/src/main/scala/com/crobox/clickhouse/dsl/misc/DSLImprovements.scala
aShevc: When compiled I have got the warning saying that the only case in which this would be captured, it is if the matched variable was null, so I replaced it, can replace back and see how it compiles in CI/CD

dsl/src/main/scala/com/crobox/clickhouse/dsl/OperationalQuery.scala
```diff
@@ -72,8 +72,8 @@ trait OperationalQuery extends Query {
 
   def groupBy(columns: Column*): OperationalQuery = {
     val internalGroupBy = internalQuery.groupBy.getOrElse(GroupByQuery())
-    val newGroupBy      = Some(internalGroupBy.copy(usingColumns = internalGroupBy.usingColumns ++ columns))
```
lwolters, line 75: Are we also using the scala formatting plugin here? If not, let's add it.

dsl/src/main/scala/com/crobox/clickhouse/dsl/Query.scala
```diff
@@ -3,7 +3,7 @@ package com.crobox.clickhouse.dsl
 import scala.util.Try
 
 trait Table {
-  val database: String
+  def database: String = ""
```
lwolters, line 6: Just curious, why did you instantiate this to an empty string? In the previous scenario compile errors would popup if not set. Now there might be a chance that an empty database is selected (which should be avoided imho)

dsl/src/main/scala/com/crobox/clickhouse/dsl/column/ComparisonFunctions.scala
```diff
@@ -2,26 +2,20 @@ package com.crobox.clickhouse.dsl.column
...
-    def >=(other: M): ExpressionColumn[Boolean] = ComparisonColumn(self, ">=", other)
-  }
-
-  def _equals(col1: ConstOrColMagnet[_], col2: ConstOrColMagnet[_]): ExpressionColumn[Boolean] = ComparisonColumn(col1 , "=", col2)
-  def notEquals(col1: ConstOrColMagnet[_], col2: ConstOrColMagnet[_]): ExpressionColumn[Boolean] = ComparisonColumn(col1 , "!=", col2)
-  def less(col1: ConstOrColMagnet[_], col2: ConstOrColMagnet[_]): ExpressionColumn[Boolean] = ComparisonColumn(col1 , "<", col2)
-  def greater(col1: ConstOrColMagnet[_], col2: ConstOrColMagnet[_]): ExpressionColumn[Boolean] = ComparisonColumn(col1 , ">", col2)
-  def lessOrEquals(col1: ConstOrColMagnet[_], col2: ConstOrColMagnet[_]): ExpressionColumn[Boolean] = ComparisonColumn(col1 , "<=", col2)
-  def greaterOrEquals(col1: ConstOrColMagnet[_], col2: ConstOrColMagnet[_]): ExpressionColumn[Boolean] = ComparisonColumn(col1 , ">=", col2)
+  def _equals(col1: ConstOrColMagnet[_], col2: ConstOrColMagnet[_]): ExpressionColumn[Boolean] = ComparisonColumn(col1, "=", col2)
```
lwolters, line 10: I really don't like the _ (which was already there btw). Can't we get rid of it? Or make this one deprecated and add the 'equals()' method. If compiler starts complaining, perhaps rename it to: 'isEquals' or something similar?

dsl/src/main/scala/com/crobox/clickhouse/dsl/column/Magnets.scala
```diff
@@ -28,6 +31,37 @@ trait Magnets {
     val column: TableColumn[C]
   }
 
+  // ComparabeWith trait and Cast case class were members of ComparisonFunctions and TypeCastFunctions trait
+  // respectively. But placing them in the mixin traits causes Scala 3 compiler to crash. Hence, placing these
+  // constructs here is a workaround allowing for the codebase to be compiled with Scala 3.
+  trait ComparableWith[M <: Magnet[_]] {
```
lwolters, line 37: In scala 3 the [_] should be rewritten to [?] right?

dsl/src/main/scala/com/crobox/clickhouse/dsl/column/Magnets.scala
```diff
@@ -120,7 +154,7 @@ trait Magnets {
       with StringSearchOps
       with EmptyNonEmptyCol[C]
 
-  implicit def stringColMagnetFromString[T <: String: QueryValue](s: T): StringColMagnet[String] =
+  implicit def stringColMagnetFromString[T <: String : QueryValue](s: T): StringColMagnet[String] =
```
lwolters, line 157: The space between String and QueryValue doesn't seem right. Looks like formatting issues

lwolters (approved):
Overall looks good. Some minor code formatting related remarks

dsl/src/main/scala/com/crobox/clickhouse/dsl/column/Magnets.scala
aShevc: @lwolters Eventually in the future, yes. See https://dotty.epfl.ch/docs/reference/changed-features/wildcards.html but for now it would be too big of a hassle to change

PR #162, opened 2024-05-24
Added small improvement

lwolters:
Small improvement. Although everything works fine, 'printing' SQL queries with a splitByChar using special ASCII characters result in unreadable code, i.e. ```splitByChar('',`abcd`)])```  This makes it impossible to copy 'n paste.

A better solution is to use the Clickhouse `char` functionality, i.e. char(31). Tested this and works


aShevc (approved):
Looks good from my side

Issue #170, opened 2024-10-11
Authentification with client ssl certificates not supported


zella:
Our clickhouse configured to secured connection with ssl certificates. Something like:
```
<config>
  <openSSL>
    <client>
      ...
      <caConfig>...</caConfig>
      <certificateFile>...</certificateFile>
      <privateKeyFile>...</privateKeyFile>
      <verificationMode>strict</verificationMode>
    </client>
  </openSSL>
<config>
```

This can be done with `HttpsConnectionContext `.  https://pekko.apache.org/docs/pekko-http/current/client-side/client-https-support.html But there no ability to pass context to `Http.superPool(...)`.  This code hardcoded in `ClickHouseExecutor`.  For example elastic client have ability to configure ssl context. https://github.com/Philippus/elastic4s/issues/2712

sjoerdmulder:
Thanks for the issue, we are open for PR's if you feel like it.

Op vr 11 okt 2024 14:27 schreef Andrey Zelyaev ***@***.***>:

zella:
@sjoerdmulder we open mr. pls check it

sjoerdmulder:
@zella thank you for your patience. Sorry it took so long to release. Version 1.2.7 is out now with the changes

PR #171, opened 2024-11-06
Add https support

antonstamov:
Add https support requested in https://github.com/crobox/clickhouse-scala-client/issues/170


lwolters:
Hi @antonstamov Thanks for this pull request. I've quickly checked it and it looks good. Overall a good addition IMHO.
Let's await @sjoerdmulder review but from my side clear to go!

sjoerdmulder:
Yes, the only thing, I don't think we need to add the `certs` as code in the repository right? e.g. they should be in `.gitignore` this so we don't have private key (although just test) stored. Just need to make sure the generate-keys.sh is called before the test.

antonstamov:
@sjoerdmulder Yes, you're right. Fixed.

maxcom:
@sjoerdmulder please review!

sjoerdmulder (approved):

sjoerdmulder:
@antonstamov sorry for the delay

maxcom:
@sjoerdmulder please release new version with this feature

sjoerdmulder:
Sorry for the delay, I tried but the release fails on the new ssl tests,
and then I gave up. Need to find some time for it

Op ma 25 nov 2024 13:59 schreef Maxim Valyanskiy ***@***.***>:

PR #206, opened 2025-06-11
DISTINCT ON support within the OperationalQuery.

okanozeren:
- DISTINCT ON support within the OperationalQuery.

### DIFF
--- a/dsl/src/main/scala/com/crobox/clickhouse/dsl/OperationalQuery.scala
+++ b/dsl/src/main/scala/com/crobox/clickhouse/dsl/OperationalQuery.scala
@@ -23,8 +23,9 @@ trait OperationalQuery extends Query {
   }
 
   def distinctOn(distinctColumns: Column*)(columns: Column*): OperationalQuery = {
-    val distinctExpr = s"DISTINCT ON (${distinctColumns.map(_.name).mkString(", ")})"
-    val newSelect    = Some(SelectQuery(columns, distinctExpr))
+    val escapedColumns = distinctColumns.map(c => escapeIdentifier(c.name))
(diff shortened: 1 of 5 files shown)


dsl/src/main/scala/com/crobox/clickhouse/dsl/OperationalQuery.scala
```diff
@@ -22,6 +22,12 @@ trait OperationalQuery extends Query {
     OperationalQuery(internalQuery.copy(select = newSelect))
   }
 
+  def distinctOn(distinctColumns: Column*)(columns: Column*): OperationalQuery = {
+    val distinctExpr = s"DISTINCT ON (${distinctColumns.map(_.name).mkString(", ")})"
```
sjoerdmulder, line 26: Can you double check if `.name` escapes SQL Injections?

okanozeren: addressed here: https://github.com/crobox/clickhouse-scala-client/pull/206/commits/2bb595d8be56bd2acef4fecfd11b5d5fd67d1f28

dsl/src/main/scala/com/crobox/clickhouse/dsl/Query.scala
```diff
@@ -17,6 +17,13 @@ trait Table {
 
 trait Query {
   val internalQuery: InternalQuery
+
+  def escapeIdentifier(identifier: String): String = {
```
sjoerdmulder, line 21: Please keep stuff DRY and use the existing function

sjoerdmulder (approved):

PR #292, opened 2026-05-26
Support expression columns in JOIN ON

toztemel:
## Summary

Allows `arrayJoin(...)` and other expression columns on either side of `JoinQuery.on`, and adds `leftCol` / `rightCol` to qualify same-named columns.

Previously the tokenizer prefixed each ON operand with `L<n>.` / `R<n>.` using `col.name`, which collapsed `ExpressionColumn` to `L1.NULL` (since `EmptyColumn.name = "NULL"`). The tokenizer now branches per kind: plain `TableColumn` keeps the auto-prefix, `ExpressionColumn` is emitted verbatim, and `JoinSideQualified` resolves to the current join's L/R alias (respecting custom from-aliases).

## API additions

- `leftCol[V](col: TableColumn[V])` / `rightCol[V](col: TableColumn[V])` in `package object dsl` — qualify a column with the LEFT/RIGHT side of the enclosing `JoinQuery.on`.
- `case class JoinSideQualified[V](side: JoinSide, inner: TableColumn[V]) extends TableColumn[V](inner.name)` — the type produced by `leftCol` / `rightCol`. Preserves `V` so it composes with magnets (e.g., `arrayJoin(leftCol(ref[Seq[String]](...)))`).
- `JoinSideQualified` is handled in `tokenizeColumn` and a new `tokenizeJoinSide` helper.
- `JoinCondition.apply(column: Column)` rejects `JoinSideQualified` (single-arg form would imply LEFT = LEFT or RIGHT = RIGHT — never user intent).
- `tokenizeJoinKeys`' USING branch rejects `JoinSideQualified` (only valid in ON).

### `arrayJoin` on the LEFT side of ON

```scala
select(dsl.all)
  .from(OneTestTable)
  .join(AllLeftJoin, TwoTestTable)
  .on((arrayJoin(leftCol(numbers)), "=", col2))
```

```sql
SELECT * FROM one_table AS L1
ALL LEFT JOIN (SELECT * FROM two_table) AS R1
ON arrayJoin(L1.numbers) = R1.column_2
```

### Same-named columns on both sides

```scala
select(dsl.all)
  .from(OneTestTable)
  .join(InnerJoin, TwoTestTable)
  .on((leftCol(itemId), "=", rightCol(itemId)))
```

```sql
SELECT * FROM one_table AS L1
INNER JOIN (SELECT * FROM two_table) AS R1
ON L1.item_id = R1.item_id
```

### Custom from-alias honoured by `leftCol`

```scala
select(dsl.all)
  .from(OneTestTable).as("ott_alias")
  .join(InnerJoin, TwoTestTable)
  .on((leftCol(itemId), "=", rightCol(itemId)))
```

```sql
SELECT * FROM one_table AS ott_alias
INNER JOIN (SELECT * FROM two_table) AS R1
ON ott_alias.item_id = R1.item_id
```

## Follow-ups

> [!WARNING]
> IT (`JoinExpressionOnIT`) is gated by `assumeMinimalClickhouseVersion(24)` and is canceled on the repo's default CH 23.8 image — `arrayJoin` in `JOIN ON` errors with `INVALID_JOIN_ON_EXPRESSION (Code 403)` on 23.8 but works on 24.8+ (verified manually). Bumping `CLICKHOUSE_VERSION` in `.docker/docker-compose.yml` requires resolving the 24+ `default` user password change, out of scope here.

> [!NOTE]
> Audit: all sibling tokenizers in `dsl/.../language/` pattern-match on specific function case classes and recurse via `tokenizeColumn(...)`, so `JoinSideQualified` flows through transparently. No other file needed changes.

## References

- https://clickhouse.com/docs/sql-reference/functions/array-join
- https://clickhouse.com/docs/sql-reference/statements/select/join

### DIFF
--- a/dsl/src/it/scala/com/crobox/clickhouse/dsl/JoinExpressionOnIT.scala
+++ b/dsl/src/it/scala/com/crobox/clickhouse/dsl/JoinExpressionOnIT.scala
@@ -1,0 +1,42 @@
+package com.crobox.clickhouse.dsl
+
+import com.crobox.clickhouse.DslITSpec
+import com.crobox.clickhouse.DslITSpec.IntResult
+import com.crobox.clickhouse.dsl.JoinQuery.{AllLeftJoin, InnerJoin}
+
+import java.util.UUID
+
+class JoinExpressionOnIT extends DslITSpec {
+
+  override val table1Entries: Seq[Table1Entry] = Seq(
+    Table1Entry(UUID.randomUUID(), numbers = Seq(1, 2, 3))
+  )
+  override val table2Entries: Seq[Table2Entry] = Seq(
+    Table2Entry(UUID.randomUUID(), "a", 1, "x", None),
+    Table2Entry(UUID.randomUUID(), "b", 3, "y", None)
+  )
+
+  // ClickHouse < 24 rejects `arrayJoin(...)` in `JOIN ON` with INVALID_JOIN_ON_EXPRESSION (Code 403).
+  it should "join with arrayJoin(leftCol(...)) on an expression LHS (INNER JOIN)" in {
+    assumeMinimalClickhouseVersion(24)
+    val query =
+      select(col2)
+        .from(OneTestTable)
+        .join(InnerJoin, TwoTestTable)
+        .on((arrayJoin(leftCol(numbers)), "=", col2))
+    val resultRows = queryExecutor.execute[IntResult](query).futureValue.rows
+    resultRows.length shouldBe 2
+  }
+
+  // ALL LEFT JOIN keeps unmatched arrayJoin'd left rows → 3 (matches 1 and 3, plus unmatched 2).
+  it should "join with arrayJoin(leftCol(...)) on an expression LHS (ALL LEFT JOIN)" in {
+    assumeMinimalClickhouseVersion(24)
+    val query =
+      select(col2)
+        .from(OneTestTable)
+        .join(AllLeftJoin, TwoTestTable)
+        .on((arrayJoin(leftCol(numbers)), "=", col2))
+    val resultRows = queryExecutor.execute[IntResult](query).futureValue.rows
+    resultRows.length shouldBe 3
+  }
+}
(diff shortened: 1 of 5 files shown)


toztemel:
## Equivalent SQL formulations

Three ways to express the same effect:

**1. Nested subquery (the current workaround this PR replaces):**

```sql
SELECT *
FROM (SELECT arrayJoin(numbers) AS n, * FROM one_table) AS L1
ALL LEFT JOIN (SELECT * FROM two_table) AS R1
  ON L1.n = R1.column_2
```

Unfold the array in an inner SELECT, expose the alias, then equi-join. Works on all ClickHouse versions but adds a subquery layer.

**2. `ARRAY JOIN` clause (FROM-level row expansion):**

```sql
SELECT *
FROM one_table AS L1
ARRAY JOIN L1.numbers AS n
ALL LEFT JOIN (SELECT * FROM two_table) AS R1
  ON n = R1.column_2
```

Unfolds the array in the FROM clause before the join. Most idiomatic for this pattern. This is the **`ARRAY JOIN` clause** (statement-level), not the `arrayJoin` function.

> **Not yet expressible in this DSL.**

**3. `arrayJoin` function in `ON` (this PR):**

```sql
SELECT *
FROM one_table AS L1
ALL LEFT JOIN (SELECT * FROM two_table) AS R1
  ON arrayJoin(L1.numbers) = R1.column_2
```

The most compact form. Requires CH 24+ (CH 23.8 rejects with `INVALID_JOIN_ON_EXPRESSION`).

All three produce the same row count for the IT fixture (`numbers=[1,2,3]` with matches on 1 and 3 → 3 rows for ALL LEFT JOIN, 2 rows for INNER JOIN).

toztemel:
Closing — the platform shipped catalogue groupBys via a nested-subquery shape (CRO-254) that works on current ClickHouse 23.8 through the existing DSL, so this fix no longer unblocks anything. Independently, `arrayJoin` in JOIN ON requires ClickHouse 24+ (`INVALID_JOIN_ON_EXPRESSION` on 23.8), so this would stay unusable until a CH bump regardless. Rationale and revisit conditions recorded on Linear CRO-255; the branch stays available if it's ever resurrected.

PR #324, opened 2026-08-21
Fix client reliability defects

### LINKED ISSUE #144: Clickhouse Internal Error when data contains "DB::Exception"
Our data can contains "DB::Exception". For example - search history in google.

When query this data we have:
`
com.crobox.clickhouse.ClickhouseException: Clickhouse Internal Error`

Maybe it there a better way to handle errors in clickhouse?

sjoerdmulder:
Targets **`1.x`** for a 1.3.1, then forward-merges to `main`. These are the client-side items deferred while the ClickHouse version work went through. All internal — nothing removed from the API.

## Progress subscribers leaked, one per request

`executeRequestInternal` called `progressSource.runForeach` on **every attempt**. That materialises a `BroadcastHub` consumer which nothing ever cancels — so one per request, plus one per retry, each retained for the lifetime of the hub, each filtering every progress event for the rest of the process.

Progress is now subscribed once per *query*, via a `KillSwitch` shut down when the query settles, alongside the existing `progressQueue.complete()`.

## Streamed failures were detected by substring

`content.contains("DB::Exception")`, carrying a FIXME acknowledging that a result row containing that text would be misread as a failure.

I checked whether `X-ClickHouse-Exception-Code` could replace it. **It cannot** — verified against 25.3 that a query failing part way through its result returns:

```
HTTP/1.1 200 OK          <- already sent, cannot be retracted
(no X-ClickHouse-Exception-Code)
{"x":0}{"x":0}...{"exception":"Code: 395. DB::Exception: boom ..."}
```

Once headers have flushed ClickHouse has no way to signal failure except appending to the body. Before they flush, the status is already 4xx/5xx and the other branch handles it — where the header *is* present.

So the body still has to be read, but it is now matched on the shape ClickHouse actually appends — `Code: <n>. DB::Exception` — which closes the false positive and yields the error code for the exception message. New unit tests cover both directions, including the rows-that-mention-it case that used to trip.

## The version probe blocked construction

`serverVersion` was a strict `val` running `Await.result(query("select version()"), 5.seconds)`. Constructing a client stalled the calling thread for up to five seconds and could not be done at all without a reachable server. Now `lazy`, so the block moves to first use — where a caller is asking for the version anyway.

Its fallback was `ClickhouseServerVersion.latest = 22.8.15`, which is neither the latest nor, since #320, a supported version. Renamed to `fallback` (what it always was), set to the oldest LTS we test against, and `latest` **deprecated rather than removed**.

Worth being precise about the impact: only one place gates on the version today — `EmptyFunctionTokenizer`, at 21.8 — and both the old and new fallback sit above it, so **no generated SQL changes**. This removes a trap for the next version gate added above 22.8 rather than fixing a live bug. I'd overstated this earlier.

## Sub-second durations truncated to zero

`getDuration(...).getSeconds.seconds` discarded everything below a second. A health-check interval of `500 millis` became `0` — a tight polling loop — and a sub-second indexer `flush-interval` likewise. Four sites now read `toMillis`.

## Smaller ones

- An unrecognised `connection.type` threw `MatchError` with nothing naming the bad value or the valid ones.
- Cluster discovery hardcoded port 8123 for every discovered host, ignoring a configured port. `system.clusters` reports the *native* port, so the only sensible assumption is that peers expose HTTP where the configured host does.
- `MultiHostBalancer` hardcoded a 5-second ask timeout while `ClusterAwareHostBalancer` used `host-retrieval-timeout` (default **one** second) for the same ask. Both now read that setting, falling back to the old 5 seconds when absent so a balancer built on a bare `ActorSystem` still works.
- `QueryRetry.retryNumber` counted against the *configured* retry total rather than the effective one, so a query overriding `settings.retries` reported the wrong attempt number.

## Not included

No per-query timeout yet. There is still none anywhere — everything inherits pekko-http pool defaults. It needs a new config key and a decision on whether to pair it with ClickHouse's `max_execution_time`, which makes it a feature rather than a fix, so it belongs in 1.4.0.

## Verification

218 dsl tests on Scala 2.13.18 and 3.3.8, 5 new parser tests, scalafmt clean. Locally the two client failures are the pre-existing environmental ones — the SSL test needs `.docker/certs`, and `ClusterConnectionFlowTest` needs the `test_shard_localhost` cluster that `.docker/server-config.xml` defines (modern ClickHouse no longer ships it by default). Both are provided by the compose stack CI uses, so CI here is the real check.

---

Fixes #144 — a 2023 report of exactly this false positive (*"Our data can contains 'DB::Exception'. For example - search history in google."*). The regex now requires the `Code: <n>. DB::Exception` shape ClickHouse actually appends, so a result row that merely contains the phrase is no longer read as a failure.

### DIFF
--- a/client/src/main/scala/com/crobox/clickhouse/ClickhouseClient.scala
+++ b/client/src/main/scala/com/crobox/clickhouse/ClickhouseClient.scala
@@ -133,7 +133,12 @@ class ClickhouseClient(
     executeRequestInternal(hostBalancer.nextHost, sql, queryIdentifier, settings, Option(entity), None)
   }
 
-  val serverVersion: ClickhouseServerVersion =
+  /**
+   * Resolved on first use, not at construction. This blocks on a query to the server, and doing that eagerly meant a
+   * client could not be constructed without a reachable ClickHouse -- it stalled the constructing thread for up to five
+   * seconds and then carried on with a guessed version. Callers that want it resolved up front can still touch it.
+   */
+  lazy val serverVersion: ClickhouseServerVersion =
     try {
       val path    = "crobox.clickhouse.server.version"
       val cfg     = configuration.getOrElse(ConfigFactory.load())
@@ -156,8 +161,12 @@ class ClickhouseClient(
       version
     } catch {
       case x: Throwable =>
-        val latest = ClickhouseServerVersion.latest
-        logger.error(s"Can't determine Clickhouse Server Version. Falling back to: $latest. Error: ${x.getMessage}", x)
-        latest
+        val assumed = ClickhouseServerVersion.fallback
+        logger.error(
+          s"Can't determine Clickhouse Server Version. Assuming: $assumed. Set `crobox.clickhouse.server.version` to " +
+            s"avoid guessing. Error: ${x.getMessage}",
+          x
+        )
+        assumed
     }
 }
--- a/client/src/main/scala/com/crobox/clickhouse/ClickhouseServerVersion.scala
+++ b/client/src/main/scala/com/crobox/clickhouse/ClickhouseServerVersion.scala
@@ -15,5 +15,13 @@ object ClickhouseServerVersion {
   def apply(version: String): ClickhouseServerVersion =
     ClickhouseServerVersion(version.split('.').toSeq.map(_.filter(_.isDigit)).filter(_.trim.nonEmpty).map(_.toInt))
 
-  def latest: ClickhouseServerVersion = ClickhouseServerVersion(versions = Seq(22, 8, 15))
+  /**
+   * Assumed when the server's version cannot be determined. Not actually the latest release -- it is a floor, chosen as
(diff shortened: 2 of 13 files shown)


sjoerdmulder:
Good catch on the `ClickhouseHostHealth` duration — that was a real bug I introduced, and worse than the one it replaced. Correcting `getDuration(...).getSeconds.seconds` across four sites, I replaced `.getSeconds` but the trailing `.seconds` sat on the next line, so it became `.toMillis.seconds`. Default 10s interval → ~2h46m, which would have silently stopped health checking rather than fixing sub-second intervals.

Also right that nothing pinned the value, so it sailed through a green local suite.

Fixed in a68de5c, structurally rather than in place: four hand-written copies of one expression is how they drifted apart, so it now lives in `ConfigDuration` and all four sites call it. Added `ConfigDurationTest` pinning the numbers — including an explicit "does not inflate whole seconds by a thousandfold" case, so this exact regression can't recur.

Everything else in the review matches my own reading, including the point that the `serverVersion` fallback change alters no generated SQL today.

Issue #329, opened 2026-08-21
Support modern ClickHouse types: DateTime64, Variant, Dynamic, JSON, SimpleAggregateFunction


sjoerdmulder:
`dsl/.../schemabuilder/ColumnType` has no representation for types ClickHouse has added since:

- **`DateTime64`** — currently unrepresentable. `DateTimeQueryValue` formats `"yyyy-MM-dd HH:mm:ss"` with no zone handling and no sub-second precision, so this is coupled to the joda → java.time migration (#326).
- **`Variant`**
- **`Dynamic`**
- **The new `JSON` type** (the native one, not the `visitParam*` functions the DSL wraps today)
- **`SimpleAggregateFunction`**

Each needs a `ColumnType` case, a tokenizer arm, and — for anything readable — a `ColumnDecoder` instance once #325 lands.

PR #352, opened 2026-08-25
Migrate joda-time to java.time

### LINKED ISSUE #326: Migrate joda-time to java.time
29 files import `org.joda`; **zero** use `java.time`. It is not an internal detail — it is in the public API of both modules, so every consumer inherits it. joda-time has been in maintenance mode with an upstream recommendation to migrate since Java 8.

### Sites

- `client/.../time/` — `MultiInterval`, `Duration`, `TimeUnit`, `package.scala`. `MultiInterval` extends joda's `BaseInterval`, so it needs a hand-rolled replacement.
- `client/.../partitioning/PartitionDateFormatter.scala`
- `dsl/.../marshalling/QueryValue.scala` — `DateTimeQueryValue`, `LocalDateQueryValue`
- `dsl/.../marshalling/ClickhouseJsonSupport.scala` — `ClickhouseIntervalStartFormat`, five regexes handling ClickHouse's relative-month encoding and ms/nano timestamps
- `dsl/.../column/Magnets.scala` — `DateOrDateTime` and its four implicits
- `dsl/.../column/{DateTimeFunctions,TypeCastFunctions,ArithmeticFunctions,ArrayFunctions,DictionaryFunctions}.scala`
- `dsl/.../misc/DateConditions.scala`, `dsl/.../schemabuilder/Engine.scala`, `dsl/.../language/ClickhouseTokenizerModule.scala`

### Mapping

`DateTime` → `ZonedDateTime` (ClickHouse `DateTime` is zone-aware), `LocalDate` → `java.time.LocalDate`, `Period`/`Duration` → `java.time` equivalents.

While here: add sub-second support so `DateTime64` becomes representable. `DateTimeQueryValue` currently formats `"yyyy-MM-dd HH:mm:ss"` with no zone handling and no sub-second precision.

sjoerdmulder:
Closes #326. 33 files, and **joda-time is gone from `build.sbt`**.

## Type mapping

| Before | After |
|---|---|
| `org.joda.time.DateTime` | `java.time.ZonedDateTime` (ClickHouse `DateTime` is zone-aware) |
| `org.joda.time.LocalDate` | `java.time.LocalDate` |
| `IntervalStart` | `ZonedDateTime` |
| joda `Period` on `MultiTimeUnit` | `(amount: Int, chronoUnit: ChronoUnit)` |
| joda `Interval` | `TimeInterval`, a two-field case class |
| `MultiInterval extends BaseInterval` | plain case class exposing `start` / `end` |

**Why not a `Period`:** java.time has no single type covering joda's. It splits into a date-based `Period` and a time-based `Duration`, and **normalises weeks into days** — `Period.ofWeeks(1).toString == "P7D"`, so `P1W` and `P7D` become indistinguishable, which `TimeUnit.apply(Period)` depended on. Carrying `(amount, ChronoUnit)` instead keeps `Week` and `7 × Day` distinct, and collapses the nine-case `nextStartFromDate` into a single `plus` call that is provably identical for every unit. `TimeUnit.apply(Period)` becomes `apply(amount, ChronoUnit)`; it had no caller in the repo.

## The alignment arithmetic is ported, not rewritten

`MultiInterval`'s epoch-millis division with an explicit zone offset — including the week-1-since-epoch adjustment — is carried over as it stood. java.time offers tidier primitives, but using them would have changed results, and #351 is what makes the difference visible: **all fifteen `MultiIntervalTest` cases pass unchanged**, including the eleven zone-sensitive ones. Local-midnight day and week alignment, 23- and 25-hour DST days, epoch-aligned sub-day buckets, Monday weeks — all identical.

That is the whole argument for having done #351 first. Without it this PR would be untestable in the ways that matter.

## Three deliberate behaviour changes

- **`PartitionDateFormatter.dateFormat(ZonedDateTime)` uses the value's own zone.** It previously went through the epoch millis and so used the system default, discarding the zone it had just been handed. This is data-affecting — it decides which partition a row lands in — so it's called out rather than buried. The zoneless overloads (`Long`, `Date`) still use the system default, since there is nothing else to use.
- **`determineZoneId` sorts the candidate ids.** A `ZonedDateTime` on `ZoneOffset.UTC` has id `"Z"`, which is *not* in `getAvailableZoneIds`, so it falls through to the `Etc/` search — where eight zones sit at offset zero and the set is unordered. Without sorting, the emitted timezone would vary between runs for a UTC interval. joda never hit this because its UTC zone id is `"UTC"`, which was in its own id list. The existing `Etc/GMT-2` assertion still passes.
- **`DateTime64` keeps sub-millisecond precision.** joda truncated anything finer than `DateTime64(3)`; `ColumnDecoder`'s scaladoc said so and pointed at this issue. There's now a test for `.250000001`.

## Deliberately not changed

`DateTimeQueryValue` still renders second precision. `ZonedDateTime` makes sub-second literals expressible, but widening the format would change **every** `DateTime` value the DSL emits, so that belongs with #329 rather than here.

## Verification

- **309 dsl unit tests** and **138 integration tests** pass; client green apart from the two known environmental failures (missing local keystore, non-clustered server).
- `grep -rn joda` over sources and `build.sbt` returns only three comments that reference it historically.

Two test-side consequences worth knowing for anyone porting downstream: `ZonedDateTime.now` carries nanoseconds where joda's carried milliseconds, so a millis round-trip no longer returns an equal value; and `getDayOfWeek` returns an enum rather than an `Int` (`.getValue` restores joda's 1–7 numbering).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -40,8 +40,7 @@ lazy val client: Project = (project in file("client"))
       "org.apache.pekko"           %% "pekko-stream"  % PekkoVersion,
       "org.apache.pekko"           %% "pekko-http"    % PekkoHttpVersion,
       "com.typesafe.scala-logging" %% "scala-logging" % "3.9.6",
-      "com.typesafe"                % "config"        % "1.4.9",
-      "joda-time"                   % "joda-time"     % "2.14.3"
+      "com.typesafe"                % "config"        % "1.4.9"
     ) ++ Seq("org.apache.pekko" %% "pekko-testkit" % PekkoVersion % Test) ++ Build.testDependencies.map(_ % Test)
   )
 
--- a/client/src/main/scala/com/crobox/clickhouse/partitioning/PartitionDateFormatter.scala
+++ b/client/src/main/scala/com/crobox/clickhouse/partitioning/PartitionDateFormatter.scala
@@ -1,18 +1,22 @@
 package com.crobox.clickhouse.partitioning
 
+import java.time.format.DateTimeFormatter
+import java.time.{Instant, LocalDate, ZoneId, ZonedDateTime}
 import java.util.Date
 
-import org.joda.time.{DateTime, LocalDate}
-import org.joda.time.format.DateTimeFormat
-
 object PartitionDateFormatter {
-  private val formatter = DateTimeFormat.forPattern("yyyy-MM-dd")
+  private val formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd")
 
-  def dateFormat(ts: Long): String = formatter.print(ts)
+  /** A bare instant carries no zone, so the system default decides the date, as it did before. */
+  def dateFormat(ts: Long): String =
+    formatter.format(Instant.ofEpochMilli(ts).atZone(ZoneId.systemDefault()))
 
-  def dateFormat(dt: LocalDate): String = formatter.print(dt)
+  def dateFormat(dt: LocalDate): String = formatter.format(dt)
 
-  def dateFormat(dt: DateTime): String = dateFormat(dt.getMillis)
+  /**
+   * Uses the value's own zone. Previously this went through the epoch millis and so used the system default instead.
+   */
+  def dateFormat(dt: ZonedDateTime): String = formatter.format(dt)
 
   def dateFormat(dt: Date): String = dateFormat(dt.getTime)
 }
--- a/client/src/main/scala/com/crobox/clickhouse/time/Duration.scala
+++ b/client/src/main/scala/com/crobox/clickhouse/time/Duration.scala
@@ -1,6 +1,6 @@
 package com.crobox.clickhouse.time
 
-import org.joda.time.Period
+import java.time.ZonedDateTime
 
 sealed trait Duration {
   val unit: TimeUnit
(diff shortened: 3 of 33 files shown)


sjoerdmulder:
All three addressed in fe1c5d7.

**1. Wire-type literals in the fixtures — fixed.** You're right about the cause: my blanket `DateTime` → `ZonedDateTime` replace caught the JSON header strings, which carry *ClickHouse's* type names rather than Scala ones. Restored at all three sites. The tests did still pass, since decoding dispatches on the `NativeColumn` type parameter and not the header, which is exactly why this needed catching by eye.

**2. `standardMillis` — fixed, and thank you, this was the real one.** I verified joda 2.14.3's behaviour before touching it rather than taking it on trust:

```
second  -> Success(1000)
hour    -> Success(3600000)
day     -> Success(86400000)
week    -> Success(604800000)
month   -> Failure(UnsupportedOperationException: Cannot convert to Duration...)
year    -> Failure(UnsupportedOperationException: Cannot convert to Duration...)
```

So the finding is precisely right: a public call that used to raise now returned ~30.44 days for a month. `MultiTimeUnit` gains `hasFixedLength`, overridden to false for `Month`/`Quarter`/`Year`, and `standardMillis` throws `UnsupportedOperationException` for those again. Two tests pin both halves — exact values for second/hour/day/week, and a throw for the calendar units. `MultiInterval` is unaffected, as it aligns those by calendar field.

**3. Unused imports — trimmed** in all twelve files, down to the symbols each actually uses.

Re-verified after the changes: `scalafmtCheckAll` clean, `+compile +Test/compile +It/compile` green on 2.13 and 3.3, 309 dsl unit tests, 17 client time tests (two new), 138 integration tests. The two `client` failures are the pre-existing environmental ones — a missing local `keystore.jks` and a non-clustered server.

Issue #333, opened 2026-08-21
Drive down the SQL-validation coverage ratchet (288 AST nodes unverified)


sjoerdmulder:
`SqlValidationITSpec` verifies generated SQL by sending it to ClickHouse's own parser, and carries a ratchet: **288 of the 346 AST case classes in `dsl/column` still have no rendering check at all.** The ratchet stops that number growing; it does not shrink it.

This matters because a tokenizer arm is a string template. A wrong one compiles, and passes any hand-written `matchSQL` assertion written to match it — which is how the tuple access operator emitted an unbalanced `x.2)` for years. The harness is the only check that cannot be written to agree with a mistake.

It has already earned this: on its first run it found `avgMap` (not a ClickHouse function at all — the `-Map` combinator exists only for sum/min/max) and an incomplete `JOIN USING` fix.

### Work

Add registry entries, lowering `UnregisteredBaseline` as you go. Good order:

1. The families with the most arms and the least coverage — `ArrayFunctions` (42), `DateTimeFunctions` (41), `AggregationFunctions` (28), `MiscellaneousFunctions` (24), `URLFunctions` (22).
2. `HigherOrderFunctions` (18) — the lambda-parameter binding is where a runtime failure already lurked.
3. `TypeCastFunctions` (17), `DistanceFunctions` (15), `StringSearchFunctions` (12).

### Follow-up worth considering

Seal the family marker traits where feasible, so the compiler enumerates them and part of this becomes a compile-time check instead of a source-scanning test.

PR #356, opened 2026-08-25
Register 91 AST nodes with the SQL validation harness

sjoerdmulder:
Part of #333. `UnregisteredBaseline` goes **288 → 197**, so a third of the backlog now has its rendering checked against ClickHouse's own analyzer rather than only against hand-written expectations.

| Family | Registered |
|---|---|
| `ArrayFunctions` | 41 |
| `HigherOrderFunctions` | 18 |
| `NullableFunctions` | 8 |
| `JsonFunctions` | 7 |
| `EncodingFunctions` | 6 |
| `IPFunctions` | 5 |
| `SplitMergeFunctions` | 4 |
| `EmptyFunctions` | 2 |

Ordered as the issue suggests, starting from the families with the most arms and the least coverage. The baseline is exact, not slack — I checked the remaining count is 197 rather than merely below it.

## Three things writing them turned up

**Entries need literals, not columns, unless they carry a FROM.** `EXPLAIN QUERY TREE` resolves identifiers, so `select(isNull(col2))` fails with `UNKNOWN_IDENTIFIER` where `select(isNull(1))` passes. Everything added here is self-contained for that reason — worth knowing before adding the next batch.

**Two functions are mistyped in a way that only the server catches.** `UUIDNumToString` and `IPv6NumToString` take `FixedString(16)`, but the DSL types both arguments as `StringColMagnet`, so a plain string compiles and then fails at runtime with `Illegal type String`. Both entries now cast explicitly via `toFixedString`.

**And the mirror image is worse, so I left it alone rather than smuggle a fix in here:** `IPv6StringToNum` is declared `ExpressionColumn[Long]` and `UUIDStringToNum` `ExpressionColumn[Byte]`, where both actually return `FixedString(16)`. So the natural composition — feeding `iPv6StringToNum` into `iPv6NumToString` — does not even compile. That's a real DSL typing defect rather than a coverage gap, and deserves its own issue rather than being buried in a test PR.

**The higher-order shapes are lifted from `HigherOrderFunctionsIT`**, which already runs them for their return values. Registering them adds the analyzer's view of the lambda-parameter binding — which the issue notes is exactly where a runtime failure already lurked.

## Verification

- `scalafmtCheckAll` clean, `+It/compile` green on 2.13 and 3.3.
- 139 integration tests pass, including all 156 registered constructs going through `EXPLAIN QUERY TREE` on a live server.
- Every construct here was iterated against the server until it passed, which is the point of the harness: a wrong entry fails rather than quietly agreeing with a mistake. Several of mine did fail first — the literal-vs-column issue and both `FixedString` cases came out of that loop rather than from reading the code.

Leaves 197 for follow-up, the largest remaining being `DateTimeFunctions` (41), `AggregationFunctions` (24), `MiscellaneousFunctions` (23) and `URLFunctions` (22).

### DIFF
--- a/dsl/src/it/scala/com/crobox/clickhouse/dsl/language/SqlValidationITSpec.scala
+++ b/dsl/src/it/scala/com/crobox/clickhouse/dsl/language/SqlValidationITSpec.scala
@@ -41,6 +41,133 @@ class SqlValidationITSpec extends DslITSpec {
   private def sem(ast: String, query: OperationalQuery) = Construct(ast, query, Semantic)
   private def syn(ast: String, query: OperationalQuery) = Construct(ast, query, Syntax)
 
+  private val nullableConstructs = Seq(
+    sem("IsNull", select(isNull(1))),
+    sem("IsNullable", select(isNullable(1))),
+    sem("IsNotNull", select(isNotNull(1))),
+    sem("IsZeroOrNull", select(isZeroOrNull(1))),
+    sem("AssumeNotNull", select(assumeNotNull(1))),
+    sem("ToNullable", select(toNullable(1))),
+    sem("IfNull", select(ifNull(1, 0))),
+    sem("NullIf", select(nullIf(1, 0)))
+  )
+
+  private val jsonConstructs = Seq(
+    sem("VisitParamHas", select(visitParamHas("""{"a":1}""", "a"))),
+    sem("VisitParamExtractUInt", select(visitParamExtractUInt("""{"a":1}""", "a"))),
+    sem("VisitParamExtractInt", select(visitParamExtractInt("""{"a":-1}""", "a"))),
+    sem("VisitParamExtractFloat", select(visitParamExtractFloat("""{"a":1.5}""", "a"))),
+    sem("VisitParamExtractBool", select(visitParamExtractBool("""{"a":true}""", "a"))),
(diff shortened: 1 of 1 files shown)


sjoerdmulder:
Taken — switched to `arrayOf(arrayOf("1", "2"), arrayOf("3"))`. You're right that the previous form read like a native `scala.Array` call when it was actually the DSL's own `Array` case class winning under `import ...dsl._`. The nested `arrayOf` says the same thing without depending on the shadow, and the server accepts it identically.

For context on where it came from: I lifted that shape from the existing `ArrayFunctionsIT`, which builds it the same way — so the same note applies there, if it's worth tidying separately.

On the two observations in the summary, both of which match what I found:

- The `*StringToNum` typing defect is now **#362**, with the reproduction. Your point that this harness wouldn't catch it is exactly right and worth restating: `SqlValidationITSpec` only checks emitted SQL, so a wrong Scala-side return type is invisible to it. That's a real boundary of the coverage work — the ratchet going to zero would still not catch #362.
- The literal-vs-column constraint (`EXPLAIN QUERY TREE` resolving identifiers, so entries need literals unless they carry a FROM) is the thing I'd most want the next batch to know.

sjoerdmulder:
Correction to my previous comment: this had already merged when I pushed, so the `arrayOf` change is in #363 instead.

PR #357, opened 2026-08-25
Give magnets structural equality

### LINKED ISSUE #71: DSL v2 doesn't fully implement equality
having the following test-case you could expect that this tests passes:
```
val first = (2 === 2).and(3 === 3)
val second = (2 === 2).and(3 === 3)
first should be(second)
```
but somehow it seems object equality is not implemented fully in the new DSL v2 and fails with a

**LogicalFunction(com.crobox.clickhouse.dsl.column.Magnets$$anon$36@62a60260,And,com.crobox.clickhouse.dsl.column.Magnets$$anon$36@664f03cb) was not equal to LogicalFunction(com.crobox.clickhouse.dsl.column.Magnets$$anon$36@30c9e27,And,com.crobox.clickhouse.dsl.column.Magnets$$anon$36@72663249)**

sjoerdmulder:
Closes #71, open since 2018. The case from the issue now passes:

```scala
val first  = (2 === 2).and(3 === 3)
val second = (2 === 2).and(3 === 3)
first shouldBe second
```

## Cause

Magnets are anonymous classes minted by implicit conversion, so they had reference equality. Every AST node holding one inherited that: `LogicalFunction` and `ComparisonColumn` are case classes, so their generated `equals` was faithfully comparing *wrappers* rather than what the wrappers hold.

## Two hierarchies, not one

The first fix didn't take, which is the interesting part. `Magnet[+C]` carries a `column`, but `LogicalOpsMagnet` carries `asOption: Option[TableColumn[Boolean]]` and **does not extend `Magnet` at all** — so adding equality to `Magnet` left the issue's own example still failing, with `Magnets$$anon$21` in the message. Both hierarchies now compare and hash on their single field, which is their only state and the only part that renders.

Every magnet implementor is a trait, so nothing overrides these. Case classes in the DSL *take* magnets as fields rather than extending them.

## Knock-on for #353

This sharpens the de-duplication that #353 introduced. That change made select-list dedup compare columns **by value**, and magnets sit inside those columns — so until now it was best-effort wherever a magnet was involved: equal expressions were kept as duplicates instead of collapsed. It failed safe in that direction, but it is now exact. There's a test for it here.

That pairing is why this was worth doing right after #353 rather than leaving it as an eight-year-old curiosity.

## Verification

- 329 unit tests (+6) and 139 integration tests, all passing — no existing expectation changed, so nothing was relying on reference equality.
- Tests cover the issue's exact case, `hashCode` agreement, that differing conditions stay unequal, that a `Set` collapses equal conditions, and the select-list dedup interaction.

### DIFF
--- a/dsl/src/main/scala/com/crobox/clickhouse/dsl/column/Magnets.scala
+++ b/dsl/src/main/scala/com/crobox/clickhouse/dsl/column/Magnets.scala
@@ -30,6 +30,16 @@ trait Magnets {
    */
   trait Magnet[+C] {
     val column: TableColumn[C]
+
+    // Magnets are anonymous classes minted by implicit conversion, so without this two built from the same value
+    // compare unequal, and every AST node holding one inherits that. The wrapped column is the only state a magnet
+    // has and the only part that renders, so it is what equality is about.
+    override def equals(other: Any): Boolean = other match {
+      case that: Magnet[_] => column == that.column
+      case _               => false
+    }
+
+    override def hashCode(): Int = column.hashCode()
   }
 
   // ComparableWith trait and Cast case class were members of ComparisonFunctions and TypeCastFunctions trait
@@ -205,6 +215,15 @@ trait Magnets {
 
   sealed trait LogicalOpsMagnet extends LogicalOps {
     val asOption: Option[TableColumn[Boolean]]
+
+    // Same reasoning as Magnet: anonymous classes from implicit conversion, with the wrapped column as their only
+    // state. This hierarchy is separate because it carries an Option rather than a column.
+    override def equals(other: Any): Boolean = other match {
+      case that: LogicalOpsMagnet => asOption == that.asOption
+      case _                      => false
+    }
+
+    override def hashCode(): Int = asOption.hashCode()
 
     def isConstTrue: Boolean = asOption match {
       case Some(Const(el: Boolean)) => el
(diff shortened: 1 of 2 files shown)


sjoerdmulder:
Real bug, and mine — fixed in the latest commit. I reproduced it before changing anything rather than patching from the description:

```
in(TableA) == in(TableB)   : true    <- should be false
in(queryA) == in(queryB)   : true
in(TableA) == in(queryA)   : true    <- equal even across the two variants
in(Seq(a)) == in(Seq(b))   : false   <- unaffected, these set column to a Tuple
addColumn keeps both       : 1       <- a column silently dropped
```

The diagnosis is right on every point, including the framing that matters most: reference equality only ever *under*-collapsed, so this traded a safe false negative for an unsafe false positive. `InFuncRHMagnet` now folds in `query` and `tableRef`, matching the treatment `LogicalOpsMagnet` already had, with five regression tests including the `addColumn` case.

Two refinements to the analysis, for the record:

**The where-clause half of the consequence does not happen.** I checked, because it would have been the more serious case:

```
where(a and b)  ->  WHERE column_4 IN test.captainAmerica AS T1 AND column_4 IN test.twoTestTable AS T2
select(a, b)    ->  SELECT column_4 IN test.captainAmerica AS T1, column_4 IN test.twoTestTable AS T2
```

`and` builds a `LogicalFunction` and `select` builds a `SelectQuery` directly — neither de-duplicates. The real exposure is `SelectQuery.addColumn` and `mergeOperationalColumns`, so it reaches users through `groupBy`/`orderBy` adding columns to a projection, not through filtering.

**`OperationalQuery` is itself an anonymous class**, so comparing `query` restores reference semantics for the subquery form: two identically built subqueries still compare unequal. That is a false negative rather than a false positive, so it is the safe direction, and it is noted in the code.

I also re-ran your audit of the other magnet traits independently and agree — only `InFuncRHMagnet` carried state beyond `column`. `ArrayColMagnet`'s extra `val` is a local helper inside a single conversion, not discriminating state.

334 unit tests on both 2.13 and 3.3, 139 integration tests.

PR #358, opened 2026-08-25
Allow subtracting one date from another

### LINKED ISSUE #72: Add algebraic operations between dates
We have a number of operations which can be executed between dates and numbers, but no support to subtract two dates. This should produce a column of int as result.

sjoerdmulder:
Closes #72, open since 2018. `dateA - dateB` now compiles and renders, giving the difference the server reports as `Int32` — days for `Date`, seconds for `DateTime`.

## Why no binding existed

`+` and `-` shared a single evidence type, `AritRetType`, and the two operators are **not symmetric** for dates. Checked against 25.3:

```
toDate('2020-01-05') - toDate('2020-01-01')  ->  4                        Int32
toDate('2020-01-01') + toDate('2020-01-02')  ->  Code: 43, illegal types
```

So simply adding a `LocalDate, LocalDate, Int` binding to `AritRetType` — the obvious fix — would have made `dateA + dateB` compile into SQL the server rejects. That asymmetry is presumably why the issue sat: the type-class shape didn't have room for it.

## Shape

Subtraction gets its own `SubtractRetType`, which accepts everything addition does via an implicit lift from `AritRetType`, plus a date on each side:

```scala
implicit def subtractFromArithmetic[L, R, O](implicit ev: AritRetType[L, R, O]): SubtractRetType[L, R, O]
implicit object LocalDateDifferenceBinding extends SubtractRetType[LocalDate, LocalDate, Int]
implicit object DateTimeDifferenceBinding  extends SubtractRetType[ZonedDateTime, ZonedDateTime, Int]
```

No existing call site changes, because the lift covers every combination `AritRetType` already had — the 329 existing tests pass untouched.

**There's a test asserting `dateCol + otherCol` still fails to typecheck**, since the whole point of the separate evidence is keeping that impossible.

## Verification

- `scalafmtCheckAll` clean; `+compile +Test/compile +It/compile` green on 2.13 and 3.3 — relevant here, since implicit resolution differs between the versions.
- 334 unit tests and 139 integration tests pass.
- A `SqlValidationITSpec` entry sends the difference through the server's analyzer, so the rendering is checked rather than assumed.

### DIFF
--- a/dsl/src/it/scala/com/crobox/clickhouse/dsl/language/SqlValidationITSpec.scala
+++ b/dsl/src/it/scala/com/crobox/clickhouse/dsl/language/SqlValidationITSpec.scala
@@ -259,6 +259,13 @@ class SqlValidationITSpec extends DslITSpec {
       "date minus date",
       select(java.time.LocalDate.of(2020, 1, 5) - java.time.LocalDate.of(2020, 1, 1))
     ),
+    Construct(
+      "datetime minus datetime",
+      select(
+        java.time.ZonedDateTime.of(2020, 1, 2, 0, 0, 0, 0, java.time.ZoneOffset.UTC) -
+          java.time.ZonedDateTime.of(2020, 1, 1, 0, 0, 0, 0, java.time.ZoneOffset.UTC)
+      )
+    ),
     Construct("sample", select(shieldId).from(OneTestTable).sample(0.1), Syntax),
     Construct("sample with an offset", select(shieldId).from(OneTestTable).sample(0.1, Option(0.5)), Syntax)
   )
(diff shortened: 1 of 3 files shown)


sjoerdmulder:
Correct — `DateTimeDifferenceBinding` had no coverage at all, so the "seconds for DateTime" half of the description was unverified by anything in the diff. Fixed.

Added four unit cases mirroring what `LocalDate` had — column minus column, literal, that datetime-plus-number still works, and that `stamp + otherStamp` still fails to typecheck — plus a `SqlValidationITSpec` entry so the rendering goes through the analyzer.

And I verified the claim itself rather than just testing that it renders:

```
toDateTime('2020-01-02 00:00:00') - toDateTime('2020-01-01 00:00:00')  ->  86400   Int32
```

86400 seconds, `Int32`, so the binding's `Int` result type is right.

Worth noting the asymmetry this leaves: `Date - Date` gives days and `DateTime - DateTime` gives seconds, both typed `Int`. That's ClickHouse's own behaviour rather than something the DSL chose, but it means the unit depends on the operand type in a way the return type doesn't reveal — a reason to prefer `dateDiff` when the unit matters. Not something to fix here.

9 unit tests on both 2.13 and 3.3, 139 integration.

PR #359, opened 2026-08-25
Add window functions: OVER, WINDOW and QUALIFY

### LINKED ISSUE #327: Window functions: OVER, WINDOW, QUALIFY
There is **no representation for `OVER`, `WINDOW` or `QUALIFY` at all**. This is the largest single gap in the DSL, and window functions are how most analytical ClickHouse queries are written now.

It is also what would give `runningDifference` a real replacement. That function was removed in 1.3.0 because ClickHouse rejects it from 24.x (`DB::Exception 721`) as error-prone — its result depends on block boundaries — and the documented replacement is a window function, which this DSL cannot express.

### Work

- A `WindowFunction` AST node carrying partition-by, order-by and a frame specification.
- An `over(...)` combinator.
- A `window` / `qualify` slot on `InternalQuery` — note it currently has 11 fields and no room for these.
- Tokenizer arms, plus registry entries in `SqlValidationITSpec` so the generated SQL is verified against the server.

### Note

Adding fields to `InternalQuery` interacts with two things fixed in 1.3.0: `:+>` now uses named arguments so a new field is a compile error rather than a silent drop, and `+` validates field by field rather than reflectively. Both were changed specifically so this work is safe.

sjoerdmulder:
Closes #327 — the issue calls this "the largest single gap in the DSL", and there was no representation for any of the three keywords.

```scala
sum(col2).over(WindowSpec(partitionBy = Seq(col3), orderBy = Seq(OrderingColumn(col2))))
// sum(column_2) OVER (PARTITION BY column_3 ORDER BY column_2 ASC)
```

`over` takes three forms: a spec written in place, a `NamedWindow` defined in the query's `WINDOW` clause, or nothing at all for `OVER ()`.

| Piece | Coverage |
|---|---|
| `WindowSpec` | `PARTITION BY`, `ORDER BY`, frame |
| `FrameMode` | `ROWS`, `RANGE` |
| `FrameBound` | `UNBOUNDED PRECEDING`/`FOLLOWING`, `CURRENT ROW`, `n PRECEDING`/`FOLLOWING` |
| `WINDOW` | one or many named definitions |
| `QUALIFY` | filters on window results, repeated calls combine with `AND` |

A frame with no `end` renders the single-bound form (`ROWS UNBOUNDED PRECEDING`) rather than inventing a `BETWEEN`.

## Two things worth pointing at

**The ordering inside a spec reuses `OrderingColumn` from #350** rather than introducing a second ordering type. That keeps `WITH FILL` and window ordering from drifting apart, and means window `ORDER BY` inherited the direction handling for free.

**Clause order follows the documented skeleton** — `HAVING`, `WINDOW`, `QUALIFY`, `ORDER BY` — with a test that puts all four in one query, since that ordering is invisible until a server rejects it.

`InternalQuery` goes 15 → 17 fields; the guards from #345 caught it, as designed.

## Verification

- Every shape was checked against a live 25.3 server **before** implementing — empty window, partition, partition + order, `ROWS BETWEEN`, unbounded `RANGE`, named window, `QUALIFY`, `row_number()`, `lagInFrame`.
- **334 unit tests** (+11) and **139 integration tests**, plus 6 new `SqlValidationITSpec` entries so each shape goes through `EXPLAIN QUERY TREE`.

## Follow-up this unblocks

The issue notes `runningDifference` was removed in 1.3.0 because ClickHouse rejects it from 24.x, and that its documented replacement is a window function the DSL could not express. It can now — worth a separate issue to add the specific window helpers (`row_number`, `rank`, `lagInFrame`, `leadInFrame`) as named functions rather than leaving callers to spell them out.

### DIFF
--- a/dsl/src/it/scala/com/crobox/clickhouse/dsl/language/SqlValidationITSpec.scala
+++ b/dsl/src/it/scala/com/crobox/clickhouse/dsl/language/SqlValidationITSpec.scala
@@ -395,6 +395,46 @@ class SqlValidationITSpec extends DslITSpec {
           java.time.ZonedDateTime.of(2020, 1, 1, 0, 0, 0, 0, java.time.ZoneOffset.UTC)
       )
     ),
+    Construct("window function over an empty window", select(sum(col2).over()).from(TwoTestTable)),
+    Construct(
+      "window function over a partition and ordering",
+      select(sum(col2).over(WindowSpec(partitionBy = Seq(col3), orderBy = Seq(OrderingColumn(col2)))))
+        .from(TwoTestTable)
+    ),
+    Construct(
+      "window function with a ROWS frame",
+      select(
+        sum(col2).over(
+          WindowSpec(
+            orderBy = Seq(OrderingColumn(col2)),
+            frame = Option(WindowFrame(FrameMode.Rows, FrameBound.Preceding(1), Option(FrameBound.CurrentRow)))
+          )
+        )
+      ).from(TwoTestTable)
+    ),
+    Construct(
+      "window function with a RANGE frame",
+      select(
+        sum(col2).over(
+          WindowSpec(
+            orderBy = Seq(OrderingColumn(col2)),
+            frame = Option(
+              WindowFrame(FrameMode.Range, FrameBound.UnboundedPreceding, Option(FrameBound.UnboundedFollowing))
+            )
+          )
+        )
+      ).from(TwoTestTable)
+    ),
+    Construct(
+      "a named window", {
+        val w = NamedWindow("w", WindowSpec(partitionBy = Seq(col3)))
+        select(sum(col2).over(w)).from(TwoTestTable).window(w)
+      }
+    ),
+    Construct(
+      "qualify on a window result",
+      select(sum(col2).over() as "s").from(TwoTestTable).qualify(ref[Long]("s") > 1)
+    ),
     Construct("sample", select(shieldId).from(OneTestTable).sample(0.1), Syntax),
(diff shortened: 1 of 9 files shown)


sjoerdmulder:
Both fixed. I checked each against 25.3 first, and one of them turned out differently than described.

**`over` on any column — confirmed, fixed.** I tested what the server actually requires before `OVER`:

```
count() OVER ()      ok
any(v) OVER ()       ok
v OVER ()            Syntax error
1 OVER ()            Syntax error
toUInt32(v) OVER ()  Code: 63, aggregate function 'toUInt32' does not exist
```

So it wants an aggregate or window function — which also rules out non-aggregate *functions*, not just bare columns and literals. `Sum`, `Count`, `Avg` and the rest are all `AggregateFunction[V]`, so `over` moved there, exactly as you suggested. All three invalid forms are now compile errors with `shouldNot typeCheck` tests. The AST node keeps its `TableColumn` field — the entry point needed narrowing, not the tree.

**`WITH FILL` inside `OVER` — the conclusion is right, the reason isn't.** The server does *not* reject it:

```
SELECT v, sum(v) OVER (ORDER BY v ASC WITH FILL FROM 1 TO 5 STEP 1) AS s FROM chk.t
1  1
2  3        <- HTTP 200, no error, and no filled rows
```

It parses `WITH FILL` and silently ignores it. That's worse than a rejection, since nothing tells the caller their fill did nothing — so `WindowSpec` now `require`s the fill to be empty, rather than emitting a clause with no effect. Thank you for catching the hole; it came directly from the `OrderingColumn` reuse I'd described as a virtue in the PR body, and it is still the right reuse, just one that needed a guard.

340 unit tests on both 2.13 and 3.3, 139 integration.

Issue #325, opened 2026-08-21
Derive the query result type from the select list


sjoerdmulder:
`execute[V: JsonReader](query: Query)` has no relationship between `V` and the query. `Query` is `trait Query { val internalQuery: InternalQuery }` — no type parameter at all — and `select(columns: Column*): OperationalQuery` erases everything. The only link between a query and its decoder is a **string**: your `as "x"` alias has to match the field name in your `jsonFormat(..., "x")`, and nothing checks it at compile time or run time.

Every integration test demonstrates the cost, e.g. decoding a `uniq` (a `UInt64`) into a `String` and calling `.toInt`.

### Shape

Named tuples and match types are unavailable — we stay cross-built on 2.13 + 3.3 because `platform` is 2.13 — so:

- `ColumnDecoder[A]`, decoding one ClickHouse value to `A`. Instances must handle what is currently fudged: 64-bit integers, `Nullable` → `Option`, arrays → `Seq`, `LowCardinality` passthrough, `DateTime64`.
- `RowDecoder[R]`, assembled positionally from `ColumnDecoder`s.
- `TypedQuery[R](underlying: OperationalQuery, decoder: RowDecoder[R])` — a wrapper, **not** a change to `OperationalQuery`. Arity overloads (`select(c1: TableColumn[A], c2: TableColumn[B]): TypedQuery[(A, B)]`, up to ~12) build the column list and the decoder in lockstep. Forward the combinators that don't change the row type. This leaves all 346 case classes and the tokenizer untouched, and leaves the untyped API in place for migration.
- `QueryExecutor.execute[R](q: TypedQuery[R])`, using the carried decoder. Deprecate the `JsonReader` overload for one release.

### Switch the read format

`ClickhouseTokenizerModule` appends `FORMAT JSON` and `ClickhouseQueryExecutor` does `_.parseJson.convertTo[QueryResult[V]]` — the whole result is materialised as a `String`, then a `JsValue` tree, then `Seq[V]`, and `FORMAT JSON`'s single top-level object cannot be streamed even in principle.

`JSONCompactEachRowWithNamesAndTypes` is line-delimited (composes with the existing `client.source` framing), positional (which is what makes tuple decoding work without alias matching), and carries column types for `ColumnDecoder` to consult.

[...]

Issue #328, opened 2026-08-21
Missing SQL clauses: ARRAY JOIN, SETTINGS, WITH, TOTALS/ROLLUP/CUBE, SAMPLE


sjoerdmulder:
Clauses with no representation in `InternalQuery`, which has 11 fields today:

- **`ARRAY JOIN`** / `LEFT ARRAY JOIN`. Note `arrayJoin` exists as a *function* (`ArrayFunctions.scala`) but the clause does not.
- **Query-level `SETTINGS`** — currently only expressible via `QuerySettings` on the HTTP request, so not per-subquery.
- **`WITH TOTALS` / `ROLLUP` / `CUBE`** — check `GroupByQuery` first, there is partial support.
- **`WITH` (CTE)** — requested in #142, and there is a stale `feature/with-clause-support` branch on origin worth looking at before starting.
- **`SAMPLE`**
- **`INTERPOLATE`** (with `ORDER BY ... WITH FILL`)
- **`PASTE JOIN`**
- **`EXPLAIN`**

Each addition is four mechanical edits across two files — case class, DSL method, tokenizer arm, and a registry entry in `SqlValidationITSpec`. Only the tokenizer arm is checked by the compiler; a missing arm is a runtime `NotImplementedError`, and a *wrong* arm is caught by nothing except the validation harness.

sjoerdmulder:
All items are implemented and merged, across five PRs:

| Item | PR |
|---|---|
| Field-addition guards (prerequisite) | #345 |
| `ARRAY JOIN` / `LEFT ARRAY JOIN`, `SAMPLE`, query-level `SETTINGS`, `PASTE JOIN`, `WITH TOTALS` result row | #346 |
| `WITH` — expression, scalar-subquery and CTE forms | #347 (also closed #142) |
| `EXPLAIN` — all six kinds | #349 |
| `ORDER BY … WITH FILL`, `INTERPOLATE` | #350 |

Three corrections to this issue's original framing, for the record:

- **`WITH TOTALS`/`ROLLUP`/`CUBE` was already fully implemented**, not partial — model, DSL and tokenizer all present. The actual gap was the return path: `QueryResult` had no `totals`, so the row the server sends was parsed and discarded. Fixed in #346.
- **The `feature/with-clause-support` branch no longer existed** on origin, so the CTE work started clean.
[...]

PR #360, opened 2026-08-25
Support DateTime64, Variant, Dynamic, JSON and SimpleAggregateFunction

### LINKED ISSUE #329: Support modern ClickHouse types: DateTime64, Variant, Dynamic, JSON, SimpleAggregateFunction

sjoerdmulder:
Closes #329. None of the five had any representation in `ColumnType`.

| Type | Notes |
|---|---|
| `DateTime64(precision[, timezone])` | precision validated to 0–9, which is what the server accepts |
| `Variant(A, B, …)` | requires at least two alternatives — one is just that type |
| `Dynamic` | |
| `JSON` | the native type, not the `visitParam*` functions the DSL already wraps |
| `SimpleAggregateFunction(f, types…)` | stores the value rather than an intermediate state, unlike `AggregateFunction` |

## The decoder half is already done

The issue asks for `ColumnDecoder` instances "once #325 lands". That landed, and #352 then made the date parser read the fractional part into a `ZonedDateTime`'s nanoseconds — so `DateTime64` reads correctly today, and `Variant`/`Dynamic`/`JSON` fall through to the `JsValue` escape hatch. There's a test that inserts `2026-08-21 22:00:33.250000001` into a real `DateTime64(9)` column and reads all nine digits back.

## Validated by creating real tables

Rather than asserting on `toString`, each type creates an actual table and compares against the server's own rendering in `system.columns`. That surfaced two things worth knowing:

- **ClickHouse normalises a `Variant`'s alternatives into sorted order.** Declaring `Variant(UInt64, String)` stores `Variant(String, UInt64)`. The test pins the server's form, since fighting it would be pointless — but a user comparing declared against actual schema needs to expect it.
- **None of the five needs an experimental flag** on 25.3, which I checked before writing the IT so it wouldn't need settings plumbing.

It also caught a bug in my own test: inside `new Table { override val name = s"modern_$name" }` the interpolated `name` binds to the member being defined, not the parameter, so every table was created as `modern_null`. Renamed the parameter and left a note.

## Not included

`DateTime` with an explicit timezone. It's the same shape as `DateTime64`'s second argument and would be natural to add, but `DateTime` is currently a `case object` and giving it a parameter replaces existing public API — that deserves its own change rather than riding along here.

## Verification

- **333 unit tests** (+10 rendering, including precision bounds and nesting inside `Nullable`/`Array`/`LowCardinality`) and **146 integration tests** (+7).

### DIFF
--- a/dsl/src/it/scala/com/crobox/clickhouse/dsl/schemabuilder/ModernColumnTypeIT.scala
+++ b/dsl/src/it/scala/com/crobox/clickhouse/dsl/schemabuilder/ModernColumnTypeIT.scala
@@ -1,0 +1,78 @@
+package com.crobox.clickhouse.dsl.schemabuilder
+
+import com.crobox.clickhouse.DslITSpec
+import com.crobox.clickhouse.dsl._
+
+import java.time.ZonedDateTime
+
+/**
+ * Creates a real table for each type, because the rendering is only right if the server accepts the DDL. Verified
+ * against 25.3 that none of these needs an experimental flag.
+ */
+class ModernColumnTypeIT extends DslITSpec {
+
+  // `suffix`, not `name`: inside `new Table { override val name = ... }` a parameter called `name` is shadowed by
+  // the member being defined, which reads as null.
+  private def createWith(suffix: String, columnType: ColumnType): Unit =
+    createWith(suffix, columnType, columnType.toString)
+
+  private def createWith(suffix: String, columnType: ColumnType, expected: String): Unit = {
+    val db     = database
+    val column = NativeColumn[String]("c", columnType)
+    val create = CreateTable(
+      table = new Table {
+        override val database: String              = db
+        override val name: String                  = s"modern_$suffix"
+        override val columns: Seq[NativeColumn[_]] = Seq(column)
+      },
+      Engine.Memory
+    )
+    clickClient.execute(s"DROP TABLE IF EXISTS ${create.table.quoted}").futureValue
+    clickClient.execute(create.query).futureValue
+    // The server's own rendering of what it stored, which is the real confirmation.
+    val stored = clickClient
+      .query(s"SELECT type FROM system.columns WHERE database = '$db' AND table = 'modern_$suffix' AND name = 'c'")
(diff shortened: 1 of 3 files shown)


sjoerdmulder:
Thanks — I agree with both notes and have deliberately left both alone. Reasoning, so the decision is on the record rather than implied by silence:

**Timezone interpolated without escaping.** `s"DateTime64($precision, '$zone')"` would emit broken DDL for a zone containing a quote. I considered a `require` rejecting `'`, and decided against it: `FixedString`, `AggregateFunctionColumn` and `Nested` all interpolate developer-supplied strings the same way, so guarding this one case would be inconsistent without fixing the pattern, and a timezone with a quote in it is not a real value. If it's worth doing, it's worth doing across `ColumnType` as one change rather than on the newest case only.

**Exhaustive matches on the sealed `ColumnType`.** Right, and worth being precise about who it affects: nothing in this repo matches exhaustively on it (I checked — `TypeCastFunctionTokenizer` matches `TableColumn` subtypes, not `ColumnType`), so it's purely a downstream concern. Since 2.0 is already breaking on much larger surfaces, this belongs in the release notes as an additive-but-recompile-affecting change rather than a separate callout.

Your point about `ColumnType` only feeding DDL rendering — hence no tokenizer arm or registry entry needed — is the part I'd most want a future reader to take from this. It's the reason this PR is as small as it is, and it's the opposite of the situation for every clause PR in #328.

PR #373, opened 2026-08-26
Add set operations, join chains, OFFSET and ORDER BY modifiers

sjoerdmulder:
> Stacked on #372. Review that one first; this diff is against it.

Four clause-level gaps in one pass, because all four touch `InternalQuery` and each would otherwise be its own breaking change. They share the positional destructure in `toRawSql`, the named arguments in `:+>` and `+`, and the field inventory `InternalQueryFieldsTest` guards — doing them together means one pass over the merge semantics instead of four.

## Set operations

`unionAll: Seq[OperationalQuery]` becomes `combinations: Seq[SetOperation]`, covering `UNION ALL`, `UNION DISTINCT`, `INTERSECT`, `INTERSECT DISTINCT`, `EXCEPT` and `EXCEPT DISTINCT`.

The chain still renders flat, which only reads left to right while every step shares one precedence level. `INTERSECT` binds tighter, so `a UNION ALL b INTERSECT c` means `a UNION ALL (b INTERSECT c)` — not what a `Seq` in that order looks like. A chain mixing them is **refused** rather than silently parenthesised, because either reading is a plausible thing to have meant; nesting the `INTERSECT` in a subquery says which.

## Join chains

`join: Option[JoinQuery]` becomes `joins: Seq[JoinQuery]`, so three tables no longer need a subquery.

Every join in a chain keys off the *first* table: the left alias is claimed once and reused while the right aliases count up, giving `L1 … R1 … R2`. A join keyed off an earlier join's right-hand side still has to nest, because a `JoinCondition` names a column and not the table it came from. `using` and `on` attach to the join most recently added, so `.join(a).using(k).join(b).using(k)` reads as it looks.

## LIMIT and OFFSET

`size` becomes an `Option`, which is what makes a bare `OFFSET n` expressible — folding the offset into `LIMIT offset, size`, the only form this used to render, cannot say that. Plus `WITH TIES`.

`WITH TIES` parses only at the very end of the clause and only after the `LIMIT offset, size` spelling. Both `LIMIT size WITH TIES OFFSET offset` and `LIMIT size OFFSET offset WITH TIES` are syntax errors — the first was my initial rendering, and the validation harness caught it.

## ORDER BY modifiers

`NULLS FIRST`/`NULLS LAST` and `COLLATE`. The grammar fixes the order at `expr [ASC|DESC] [NULLS …] [COLLATE …] [WITH FILL …]` and rejects every other arrangement, so all three keyword orders were checked against a server before the tokenizer was written. `nulls` and `collate` come after `fill` in the parameter list but before it in the emitted SQL, so existing positional `OrderingColumn(col, DESC, Option(WithFill()))` call sites still compile.

## Incidental

The join chain now renders from a filtered `Seq` rather than `stripMargin`, which drops the stray double space `GLOBAL` used to emit — invisible through `matchSQL`, which normalises whitespace.

## Testing

Behavioural ITs alongside the harness entries: the harness cannot tell `UNION ALL` from `UNION DISTINCT`, a chain that joins the right tables from one that merely parses, or `NULLS FIRST` from a plain direction.

## Breaking changes

- `InternalQuery.unionAll` → `combinations`, retyped to `Seq[SetOperation]`
- `InternalQuery.join` → `joins`, retyped to `Seq[JoinQuery]`
- `Limit.size` becomes `Option[Long]`

### DIFF
--- a/dsl/src/it/scala/com/crobox/clickhouse/dsl/SetOperationAndJoinChainIT.scala
+++ b/dsl/src/it/scala/com/crobox/clickhouse/dsl/SetOperationAndJoinChainIT.scala
@@ -52,6 +52,14 @@ class SetOperationAndJoinChainIT extends DslITSpec {
   private def labels(query: OperationalQuery): Seq[String] =
     queryExecutor.executeRows(query).futureValue.rows.map(_.requiredByName[String]("column_3"))
 
+  // The discriminating case for the parenthesising of nested branches. Intended is
+  // `two UNION ALL (three EXCEPT two)` = {shared, onlyTwo, onlyThree}; a flat chain would group as
+  // `(two UNION ALL three) EXCEPT two` = {onlyThree}, which is a different set rather than a different spelling.
+  "A nested set operation" should "keep the grouping the call structure states" in {
+    ids(fromTwo.unionAll(fromThree.except(fromTwo))) should contain theSameElementsAs
+    Seq(shared, onlyTwo, onlyThree).map(_.toString)
+  }
+
   "A chain of joins" should "keep only the rows every table in the chain has" in {
     val query = select(col3)
       .from(TwoTestTable)
--- a/dsl/src/it/scala/com/crobox/clickhouse/dsl/SetOperationAndJoinChainIT.scala
+++ b/dsl/src/it/scala/com/crobox/clickhouse/dsl/SetOperationAndJoinChainIT.scala
@@ -1,0 +1,98 @@
+package com.crobox.clickhouse.dsl
+
+import com.crobox.clickhouse.DslITSpec
+
+import java.util.UUID
+
+/**
+ * Behavioural cover for the set operations and for a join chain. The validation harness proves the server accepts what
+ * these render; it cannot tell `UNION ALL` from `UNION DISTINCT`, or a chain that joins the right tables from one that
+ * happens to parse.
+ */
+class SetOperationAndJoinChainIT extends DslITSpec {
+
+  private val shared    = UUID.randomUUID()
+  private val onlyTwo   = UUID.randomUUID()
+  private val onlyThree = UUID.randomUUID()
+
+  // column_3 carries which row it is, so a join chain that keeps the wrong row shows up as the wrong value rather
+  // than only as the wrong count.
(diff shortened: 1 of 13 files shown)


sjoerdmulder:
Both findings were real. Fixed in 55630a2.

**Finding 1 — the nesting bug is confirmed, and it is broader than the framing.** Verified against 25.3 with `a={9}, b={0,1,2,3}, c={9}`:

```
SELECT 9 UNION ALL (SELECT number FROM numbers(4) EXCEPT SELECT 9)  =>  9 0 1 2 3   (intended)
SELECT 9 UNION ALL SELECT number FROM numbers(4) EXCEPT SELECT 9    =>  0 1 2 3     (what was emitted)
```

The mechanism is not precedence tiers. It bites whenever the right operand carries combinations of its own, `INTERSECT` or not, because the chain renders flat and ClickHouse reads a flat chain left to right — and set difference is not associative. Worth noting the two examples in the review both happen to coincide set-theoretically: `(a∩b)\c` and `a∩(b\c)` are the same set, so `INTERSECT`/`EXCEPT` nesting is safe either way. It is `UNION` with `EXCEPT` that actually changes the result.

So the fix keys off *the branch having its own combinations* rather than off which kinds are involved.

**Parenthesised rather than refused.** The suggestion was to reject, consistent with the flat-chain `require`. I went the other way: in the flat case the order says nothing about which grouping was meant, so refusing is the only honest answer — but here the call structure states it, and a parenthesised branch parses in every position. Refusing would reject code whose meaning is unambiguous.

**Finding 2 — `Limit(None, _, withTies = true)`** now has a `require`, matching `WithFill`.

**On the versioning point:** this is the 2.0 branch, so source- and binary-breaking changes are in scope by design — 1.3.0 is the last 1.x and there is no maintenance branch. The `InternalQuery`/`Limit` changes are already in the PR body; they will go in the release notes rather than being avoided.

Also registered `count(DISTINCT)` under a combinator in the harness, which was the untested case raised on #372: it renders `countIf(DISTINCT c, cond)`, and 25.3 accepts it and reads it as counting distinct values among the matching rows — so no guard is needed there.

Rebased onto `main` now that #372 has landed; 464 unit tests green on 2.13.18 and 3.3.8, 182 integration tests against 25.3.

PR #374, opened 2026-08-26
Open the FROM ADT for table functions

sjoerdmulder:
> Stacked on #373. Review that one first; this diff is against it.

`FromQuery` was sealed with exactly two cases, so `FROM numbers(10)` was unreachable and nothing outside the package could add it.

Opening the ADT is the part that has to happen while the API can still break — a third case breaks exhaustive matches downstream. Individual functions are additive afterwards, which is why the set below is deliberately small: the point of the PR is the third case, not the roster.

## What this adds

`TableFunctionFromQuery`, plus builders for `numbers`, `zeros`, `values`, `merge`, `remote`, `cluster`, `clusterAllReplicas` and `generateRandom` — and a generic `tableFunction(name, args*)` so anything without a wrapper, or anything the server gains later, is reachable without waiting for one.

Arguments are `Column`s, which means the existing literal rendering applies: `const("system")` becomes `'system'` and `tuple(...)` becomes a row. Functions that take a *query* rather than values, such as `view`, have no representation here — a subquery already goes in `FROM` directly.

`file`/`url`/`s3` are left for later; they need credential and format plumbing that has nothing to do with the ADT change.

## Two collisions worth naming

**`mergeTables`, not `merge`.** The `-Merge` aggregate combinator already holds that name in the same import, and an overload between the two would resolve by argument type rather than by intent.

**`join` needed its own overload.** `FromQuery extends OperationalQuery`, so without a `TableFunctionFromQuery` overload a table function took the *subquery* overload, got wrapped in an `InnerFromQuery`, and rendered as `(FROM numbers(3))` — a `SELECT`-less subquery. The tests caught it.

## FINAL and SAMPLE

Refused, for the reason a subquery refuses them: both read something the table's engine declares, and a function has no engine.

## Testing

Registered in the validation harness, with `cluster` at syntax level only since resolving one reads the server's own configuration. `TableFunctionIT` executes them, so an argument order that parses either way is still caught.

## Breaking changes

- `FromQuery` gains a third case, so exhaustive matches over it need updating.

### DIFF
--- a/dsl/src/it/scala/com/crobox/clickhouse/dsl/TableFunctionIT.scala
+++ b/dsl/src/it/scala/com/crobox/clickhouse/dsl/TableFunctionIT.scala
@@ -1,0 +1,55 @@
+package com.crobox.clickhouse.dsl
+
+import com.crobox.clickhouse.{dsl, DslITSpec}
+
+/**
+ * Table functions executed rather than only rendered, so an argument order that parses either way is still caught.
+ *
+ * Qualified as `dsl.numbers` / `dsl.all`: an inherited member shadows an imported one, and this spec inherits both a
+ * `numbers` column from the test schema and ScalaTest's `all` matcher.
+ */
+class TableFunctionIT extends DslITSpec {
+
+  private def longs(query: OperationalQuery, column: String): Seq[Long] =
+    queryExecutor.executeRows(query).futureValue.rows.map(_.requiredByName[Long](column))
+
+  "numbers" should "count up from zero" in {
+    longs(select(dsl.all) from dsl.numbers(3), "number") shouldBe Seq(0, 1, 2)
+  }
+
+  it should "count up from the offset when given one" in {
+    longs(select(dsl.all) from dsl.numbers(10, 3), "number") shouldBe Seq(10, 11, 12)
+  }
+
+  "zeros" should "produce the row count without a counter" in {
+    val rows = queryExecutor.executeRows(select(dsl.all) from dsl.zeros(3)).futureValue.rows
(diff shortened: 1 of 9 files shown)


sjoerdmulder:
Good catch, and you are right that I got this wrong. Fixed in 3493ce9.

`selectFromTable` did throw `MatchError` for a table-function `FROM`, and it is public API in the published `dsl` module, so it is exactly the regression opening the ADT could cause downstream.

Worth admitting how it slipped through: **the compiler warned about it and I dismissed the warning as pre-existing.** It was not — `git show main:...DSLImprovements.scala` has the same two-case match, but compiling `main` produces no warning for that line because `FromQuery` had only two cases there. Checking that before waving it off would have caught this.

Applied the suggested arm, plus a regression test covering all three cases (table function → `None`, table → `Some`, subquery → `None`).

I also re-audited every match on `FromQuery` in the published modules. The only other one is `insertConstraint` immediately below, which has a catch-all `case _ =>`, so a table function falls through and leaves the query unchanged — correct as-is. `parallel/package.scala` was already converted to a fold in this PR.

On the two non-blocking notes:

- **`remote()` credentials in query text** — agreed, and inherent to the table function. Not something the wrapper can avoid, and the reason `file`/`url`/`s3` were left out of this PR is the same family of concerns.
- **Downstream exhaustive matches on `FromQuery`** — agreed this belongs in the release notes, alongside the other breaking items. This is the 2.0 branch, so the break itself is in scope.

Rebased onto the updated #373; 464 unit tests green on 2.13.18 and 3.3.8, 182 integration tests against 25.3.

PR #375, opened 2026-08-26
Add the data-changing statements

sjoerdmulder:
> Stacked on #374. Review that one first; this diff is against it.

`DELETE`, `UPDATE` and `DROP PARTITION` had no representation. A caller wanting one composed the condition with the DSL, rendered it with `tokenizeColumn`, and spliced the string into hand-written SQL — which puts the escaping outside anything that checks it.

Taking the `WHERE` as an ordinary `TableColumn` is the point of the change: same DSL, same tokenizer, same escaping as a `SELECT`'s. There is a test that a value containing `' OR 1 = 1 --` is escaped rather than spliced.

## What this adds

`DeleteFrom`, `AlterDelete`, `AlterUpdate` and `DropPartition`, with a `TokenizerModule.toStatementSql` entry point and `QueryExecutor.executeStatement`.

Outside `InternalQuery`, for the reason `ExplainKind` is: these wrap a whole statement, so as a field they would render inside a subquery's parentheses and survive a query merge. One tokenizer entry point rather than four — `toExplainSql` earns its own only because its signature differs.

## Two spellings of DELETE, both here

They are not interchangeable operationally, so both are modelled. `DeleteFrom` is the lightweight delete, which marks rows deleted; `AlterDelete` is the mutation, which rewrites every part the condition touches and runs asynchronously.

## UPDATE is the ALTER TABLE spelling

Not the standalone `UPDATE t SET …`. That one does not parse at all in 25.3, the oldest release the CI matrix covers, where it fails as a syntax error. The mutation form works across the whole supported range.

## DROP PARTITION

Distinguishes the two ways to name a partition, because they render differently: an id is a quoted literal, a partitioning expression is not.

## Testing

`StatementIT` builds its own MergeTree table — the shared fixtures are `Engine.Memory`, which supports neither lightweight deletes nor mutations — and runs every statement with `mutations_sync`, since a mutation returns once queued rather than once applied.

### DIFF
--- a/dsl/src/it/scala/com/crobox/clickhouse/dsl/StatementIT.scala
+++ b/dsl/src/it/scala/com/crobox/clickhouse/dsl/StatementIT.scala
@@ -1,0 +1,118 @@
+package com.crobox.clickhouse.dsl
+
+import com.crobox.clickhouse.DslITSpec
+import com.crobox.clickhouse.dsl.schemabuilder.{ColumnType, CreateTable, Engine}
+import spray.json.DefaultJsonProtocol._
+import spray.json.RootJsonFormat
+
+/**
+ * The data-changing statements, executed. Needs its own table: the shared fixtures are Engine.Memory, which supports
+ * neither lightweight deletes nor mutations, and a DROP PARTITION needs something partitioned to drop.
+ */
+class StatementIT extends DslITSpec {
+
+  private object MutableTable extends Table {
+    override val database             = StatementIT.this.database
+    override val name                 = "mutableTestTable"
+    val part: NativeColumn[Int]       = NativeColumn("part", ColumnType.UInt32)
+    val key: NativeColumn[Int]        = NativeColumn("key", ColumnType.UInt32)
+    val payload: NativeColumn[String] = NativeColumn("payload")
+    override val columns              = List(part, key, payload)
+  }
+
+  private case class Entry(part: Int, key: Int, payload: String)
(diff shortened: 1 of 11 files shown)


sjoerdmulder:
No code change from this review — the two substantive points are accurate but resolve to release-notes items rather than fixes.

**New abstract members on public traits.** Correct: `TokenizerModule.toStatementSql` and `QueryExecutor.executeStatement` will break any downstream implementation of either trait until it adds them. That is in scope on the 2.0 branch — 1.3.0 is the last 1.x, so this release is breaking by design, and it is not going out as a patch. It belongs in the release notes with the rest of the breaking set, which is the maintainer decision you flagged rather than something to work around; giving `toStatementSql` a default is not possible since it needs `tokenizeColumn`.

**`Assignment` not tying its two types together.** Agreed it is loose, and agreed it matches the surrounding DSL. Tightening it would need the same treatment the magnets get, and `TableColumn` covariance makes an exact-match constraint hard to enforce anyway — the same problem documented on the `lagInFrame` default in #372. Left as-is deliberately rather than by oversight.

**Case-class construction rather than a fluent builder** — intended. These are single-statement ADTs with no clause ordering to accumulate, so there is nothing for a builder to add; `ExplainKind` is constructed the same way.

Note this branch also carries #376 (Decimal/IPv4/IPv6), which was merged into it. Rebased onto the updated #374, which picked up the `selectFromTable` fix; 464 unit tests green on 2.13.18 and 3.3.8, 182 integration tests against 25.3.

Issue #370, opened 2026-08-26
StackOverflowError comparing or hashing an arithmetic expression, cast, or toFixedString


sjoerdmulder:
**Blocker.** Found in RC2 consumer testing. New in 2.0 (introduced by #357, with a fifth site added by #364).

```scala
intDiv(timestampColumn, 1000).hashCode()   // StackOverflowError
abs(col2) == abs(col2)                     // StackOverflowError
cast(col2, ColumnType.UInt32).hashCode()   // StackOverflowError
toFixedString(col3, 2).hashCode()          // StackOverflowError
Seq[Column](expr).contains(expr)           // StackOverflowError
```

## Cause

#357 put `equals`/`hashCode` on `trait Magnet`, comparing the wrapped column:

```scala
override def equals(other: Any): Boolean = other match {
  case that: Magnet[_] => column == that.column
  case _               => false
}
override def hashCode(): Int = column.hashCode()
```

Four classes define their `column` as themselves:

| | |
|---|---|
| `ArithmeticFunctions.scala:15` | `ArithmeticFunctionCol` |
| `ArithmeticFunctions.scala:22` | `ArithmeticFunctionOp` |
| `Magnets.scala:80` | `Cast` |
| `TypeCastFunctions.scala:287` | `FixedString` (added in #364) |

so `equals` reads `this == that.column` and `hashCode` reads `this.hashCode()`.

### Why it wasn't obvious

Every concrete subclass is a **case class**, so the synthesised field-wise `equals` looks like it would win. It doesn't. The compiler synthesises `equals`/`hashCode` for a case class only when no concrete definition is inherited from a parent other than `AnyRef` (SLS 5.3.2) — and `Magnet` supplied one. So `IntDiv`, `Plus`, `Abs`, `Cast` and `FixedString` inherited `column == that.column` and recursed on the first call.

`Sum`, which is an `ExpressionColumn` but not a `Magnet`, kept its case-class equality (`sum(col2) == sum(col2)` is `true`) — that contrast is what pins the cause down.

### Why it surfaced only now

Nothing compared these nodes by value until `mergeOperationalColumns` began de-duplicating with `selectForGroupCols.contains(candidate)`. 1.3.0 de-duplicated by name and never called `equals` on them. RC1 had the same latent trap; its 193 alias-duplication failures (#368) masked it.

## Fix

[...]

PR #378, opened 2026-08-26
Render GROUP BY keys as expressions rather than naming them

### LINKED ISSUE #377: GROUP BY on an expression groups by the wrong thing
Found while investigating #370. **Pre-existing — v1.3.0 has byte-identical code, so grouping by an expression has never worked.**

`tokenizeGroupBy` passes its keys through `aliasOrName`, which renders an `AliasedColumn`'s alias and everything else's `name`. For a stored column that is right. For an expression it is not — `Column.name` for an `ExpressionColumn` is its **argument's** name:

```scala
abstract class ExpressionColumn[+V](targetColumn: Column) extends TableColumn[V](targetColumn.name)
```

## The dangerous case: succeeds, wrong results

```scala
select(count()).from(t).groupBy(toStartOfDay(stamp))
```
```sql
SELECT count(), toStartOfDay(ts_dt) FROM t GROUP BY ts_dt
--                                                  ^^^^^ the raw timestamp
```

The server accepts it and aggregates over the wrong buckets. Over 48 hourly rows:

| grouping | first group |
|---|---|
| `GROUP BY ts_dt` | 1 row (48 groups) |
| `GROUP BY toStartOfDay(ts_dt)` | 2 rows (3 groups) |

No error, no warning — just wrong aggregation.

## The loud case

`ArithmeticFunctionOp` is built on `EmptyColumn`, whose `name` is the string `"NULL"`:

```scala
groupBy(intDiv(timestampColumn, 1000))   // GROUP BY NULL
// Code: 215. Column ts is not under aggregate function and not in GROUP BY keys
```

## What is correct today

| key | emitted | |
|---|---|---|
| `col2` (stored column) | `GROUP BY column_2` | ✅ |
| `ref[T]("d")` | `GROUP BY d` | ✅ |
| `expr as "d"` | `GROUP BY d` | ✅ |
| `toStartOfDay(stamp)` | `GROUP BY ts_dt` | ❌ silently wrong |
| `intDiv(ts, 1000)` | `GROUP BY NULL` | ❌ rejected |
| `const(1)` | `GROUP BY NULL` | ❌ rejected |

## Fix

Render GROUP BY keys as themselves; keep referring to aliases and stored columns by name. `aliasOrName` stays as-is for `INTERPOLATE`, its only other caller, where naming a projected column is exactly what the clause wants.

Also registers the three shapes in `SqlValidationITSpec`, so what GROUP BY emits is checked against a real server from now on — this whole family was unregistered, which is how it survived.

sjoerdmulder:
Found while investigating #370. **Pre-existing** — v1.3.0 has byte-identical `aliasOrName` and `ArithmeticFunctionOp`, so grouping by an expression has never worked.

`tokenizeGroupBy` passed its keys through `aliasOrName`, which renders an `AliasedColumn`'s alias and everything else's `name`. Right for a stored column; wrong for an expression, whose `name` is its **argument's** name:

```scala
abstract class ExpressionColumn[+V](targetColumn: Column) extends TableColumn[V](targetColumn.name)
```

## The dangerous case: succeeds, wrong results

```sql
SELECT count(), toStartOfDay(ts_dt) FROM t GROUP BY ts_dt
--                                                  ^^^^^ the raw timestamp
```

Over 48 hourly rows spanning 3 days:

| grouping | result |
|---|---|
| `GROUP BY ts_dt` | 48 groups |
| `GROUP BY toStartOfDay(ts_dt)` | 3 groups |

No error — just wrong aggregation.

## The loud case

`ArithmeticFunctionOp` is built on `EmptyColumn`, whose name is the string `"NULL"`:

```
groupBy(intDiv(ts, 1000))  ->  GROUP BY NULL
Code: 215. Column ts is not under aggregate function and not in GROUP BY keys
```

## After

| key | emitted |
|---|---|
| `col2` | `GROUP BY column_2` (unchanged) |
| `ref[T]("d")` | `GROUP BY d` (unchanged) |
| `expr as "d"` | `GROUP BY d` (unchanged) |
| `toStartOfDay(stamp)` | `GROUP BY toStartOfDay(ts_dt)` |
| `intDiv(ts, 1000)` | `GROUP BY intDiv(ts, 1000)` |
| `const(1)` | `GROUP BY 1` |

`aliasOrName` is untouched and still used for `INTERPOLATE`, its only other caller, where naming a projected column is what the clause wants.

## Tests

`GroupByKeysTest` covers all six shapes above plus `WITH ROLLUP` / `WITH CUBE` / `WITH TOTALS` staying attached. The three expression shapes are registered in `SqlValidationITSpec` (now 394 constructs), so what GROUP BY emits is checked against a real server from now on — this whole family was unregistered, which is how it survived to 2.0.

- 399 dsl unit tests on 2.13.18 and 3.3.8
- 146 dsl integration tests against 25.3

Closes #377

### DIFF
--- a/dsl/src/it/scala/com/crobox/clickhouse/dsl/language/SqlValidationITSpec.scala
+++ b/dsl/src/it/scala/com/crobox/clickhouse/dsl/language/SqlValidationITSpec.scala
@@ -681,6 +681,20 @@ class SqlValidationITSpec extends DslITSpec {
       "group by a column an aggregate already covers, which is now projected",
       select(sum(col2)).from(TwoTestTable).groupBy(col2)
     ),
+    // GROUP BY named its keys rather than rendering them, so an expression key became its argument's name -- or NULL
+    // for an arithmetic one, which the server rejects outright.
+    Construct(
+      "group by a function expression",
+      select(count()).from(OneTestTable).groupBy(toStartOfDay(toDateTime(timestampColumn)))
+    ),
+    Construct(
+      "group by an arithmetic expression",
+      select(count()).from(OneTestTable).groupBy(intDiv(timestampColumn, 1000))
+    ),
+    Construct(
+      "group by a nested expression",
+      select(count()).from(OneTestTable).groupBy(toDateTime(intDiv(toUInt64rNull(shieldId), 1000)))
+    ),
     Construct(
       "order by an expression differing from the projection",
       select(sum(col2)).from(TwoTestTable).orderBy(uniq(col2))
--- a/dsl/src/main/scala/com/crobox/clickhouse/dsl/language/ClickhouseTokenizerModule.scala
+++ b/dsl/src/main/scala/com/crobox/clickhouse/dsl/language/ClickhouseTokenizerModule.scala
@@ -657,10 +657,8 @@ trait ClickhouseTokenizerModule
   private def tokenizeGroupByKeys(columns: Seq[Column])(implicit ctx: TokenizeContext): String =
     columns
       .map {
-        case EmptyColumn             => ""
+        // The alias alone. tokenizeColumn would render `<expression> AS d`, which is a projection, not a key.
         case alias: AliasedColumn[_] => alias.quoted
-        case col: NativeColumn[_]    => col.quoted
-        case col: RefColumn[_]       => col.quoted
         case col                     => tokenizeColumn(col)
       }
       .filter(_.nonEmpty)
--- a/dsl/src/main/scala/com/crobox/clickhouse/dsl/language/ClickhouseTokenizerModule.scala
+++ b/dsl/src/main/scala/com/crobox/clickhouse/dsl/language/ClickhouseTokenizerModule.scala
@@ -577,10 +577,10 @@ trait ClickhouseTokenizerModule
         s"$keyword ${removeSurroundingBrackets(tokenizeColumn(condition).trim)}"
     }
 
-  private def tokenizeGroupBy(groupBy: Option[GroupByQuery]): String = {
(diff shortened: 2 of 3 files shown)


sjoerdmulder:
Both observations addressed — one taken, one checked and found not to hold.

**Redundant arms — taken, partly.** `NativeColumn` and `RefColumn` are dropped; you're right that `tokenizeColumn`'s fallback is `col.quoted` for both, so they were restating it. `AliasedColumn` stays, because that one genuinely differs — `tokenizeColumn` renders it `<expression> AS d` (`ClickhouseTokenizerModule.scala:298`), which is a projection, not a key. Falling through would emit `GROUP BY toStartOfDay(ts_dt) AS d`. Comment now says which arm is load-bearing and why.

**Dangling `WITH ROLLUP` — checked against 25.3, and it isn't invalid:**

| query | result |
|---|---|
| `SELECT count() FROM numbers(3) WITH ROLLUP` | `3` |
| `SELECT count() FROM numbers(3) WITH TOTALS` | `3` |
| `SELECT count() FROM numbers(3) GROUP BY  WITH ROLLUP` | `Code: 62. Syntax error at position 47` |

ClickHouse accepts a trailing modifier with no `GROUP BY`. The third row is what the **old** code emitted for `groupBy(EmptyColumn).withRollup`, since `aliasOrName(EmptyColumn)` returned `""` and the clause was still wrapped. So dropping the empty clause fixes that edge rather than introducing a break — the opposite direction from the concern. Pinned as a test either way, so the question is settled in code rather than in a comment thread.

Worth noting the premise is independently pre-existing: `select(count()).from(t).withTotals` with no `groupBy` at all already emitted a bare `WITH TOTALS` before this PR, and still does. Valid, per the table above.

400 dsl unit tests on 2.13.18 and 3.3.8; 146 integration tests against 25.3.

PR #382, opened 2026-08-27
Rework higher-order array functions: own element types, one name per family, arities 1-5

sjoerdmulder:
Every multi-array higher-order function bound all its arrays to a single element type:

```scala
def arrayMap2[I, O](
    func: (TableColumn[I], TableColumn[I]) => ExpressionColumn[O],
    array1: ArrayColMagnet[_ <: Iterable[I]],
    array2: ArrayColMagnet[_ <: Iterable[I]]   // same I
)
```

So a lambda over an id array and a numeric array would not compile. The two-array case had a workaround — cast the numerics to a common type — but the **three-array case had none**, because no cast unifies a `String` with a numeric.

ClickHouse imposes no such restriction. Its own reference says a lambda *"that accepts multiple arguments can also be passed"*, over *"several arrays of identical length"* — arity is open and only the **lengths** must agree. Verified against 25.3:

```sql
SELECT arrayMap((pid, price, qty) -> concat(pid, ':', toString(price * qty)), ['a','b'], [1.5, 2.5], [2, 3])
-- ['a:3','b:7.5']
```

## What changed

**Each array carries its own element type.** The blocker, and the reason for the PR.

**One name per family.** `arrayMap2` and `arrayMap3` are both `arrayMap`, resolved by argument count. Verified on 2.13.18 and 3.3.8 before relying on it — overloads with differing type-parameter counts coexist, and explicit type arguments, annotated lambda parameters and arity one all work.

**Arities one to five.** ClickHouse takes arbitrarily many arrays; four and five were confirmed working directly against the server.

**The lambda lives in a `Lambda` ADT**, one case per arity, instead of N parallel `Option` slots on every node. Extending the old shape to five arities would have meant 911 lines, a seven-parameter base class and four `Nothing`s in every arity-one construction, with arity six needing the same surgery again. The ADT instead:

- drops the base class from five type parameters to **two**, with **eight of eighteen nodes needing none** (`ArrayAll` is one line where it was six)
- makes one more arity a case plus a rendering arm, not a change to eighteen nodes
- comes in at **861 lines** across both files, node section smaller than before
- needs **no casts** — each rendering arm matches a concrete `Lambda` case, so parameter types are recovered by the pattern rather than asserted

`Lambda` sits in `dsl`, not beside the nodes in `dsl.column`: it renders nothing on its own, and the coverage gate over that package expects everything there to. It joins `Window` and `OrderingColumn`, which are also structures rather than functions.

## Two typing corrections that fall out

`arraySort` and `arrayReverseSort` return **the first array reordered**, not the lambda's values, so they are `Iterable[I1]` and not `Iterable[O]`. Wrong at every arity, and invisible because the existing tests use one type for both — the fixture sorts `[1,3,3,0]` by `x*y` and asserts `[0,1,3,3]`, which are first-array elements, while the declared type said otherwise.

The filtering family — `arrayFilter`, `arrayFill`, `arrayReverseFill`, `arrayFirst`, `arraySplit`, `arrayReverseSplit` — is likewise keyed to the first array. Already true; now expressible rather than incidental, which is what lets `arrayFilter[String, Int]` over ids and quantities return ids.

## Known hazard

**The type-argument count selects the overload.** `arraySum[Int, Int]` picks arity one and then rejects three value arguments, where arity two wants `[Int, Int, Int]`. A compile error rather than a wrong query, and annotating the lambda's parameters avoids it entirely — but it is a real edge of the single-name scheme, and it caught me while updating the tests.

Two smaller limits, both pre-existing: lambda parameter types are not inferred, because the lambda precedes the arrays; and the lambda must return an `ExpressionColumn`, so a bare column will not serve as a sort key.

## Also confirmed from the docs

*"For some functions the first argument (the lambda function) can be omitted. In this case, identical mapping is assumed."* That is exactly the eight families whose arity-one builder takes an `Option` — a split previously reverse-engineered from the existing signatures, now documented.

The same page mentions **`arrayFold`, which this DSL lacks entirely.** Out of scope; worth its own issue.

## Breaking changes

- `arrayXxx2` / `arrayXxx3` are gone — use `arrayXxx` at any arity.
- `HigherOrderFunction` goes from three type parameters to two; most nodes lose theirs entirely.
- `arraySort` / `arrayReverseSort` result type changes from `Iterable[O]` to `Iterable[I1]`.
- `Lambda` is new public API in `com.crobox.clickhouse.dsl`.
- Explicit type applications need a count matching the intended arity (see the hazard above).

## Testing

487 unit tests green on 2.13.18 and 3.3.8, 186 integration tests against a live 25.3 server — including the previously impossible three-array expression and arities four and five, all executed end to end with results checked.


sjoerdmulder:
Reworked in 5d8da0e, answering both questions raised on the PR: **one name per family, and arities one to five.**

## One name for every arity

`arrayMap2` and `arrayMap3` are both `arrayMap` now, resolved by argument count. Verified on **2.13.18 and 3.3.8** before committing to it — overloads with differing type-parameter counts coexist, and explicit type arguments, annotated lambda parameters and arity one all still work.

This also **undoes the `arraySplit2` rename** the first commit introduced, which was the loudest break in the original PR. `arraySplit` is simply every arity again.

One hazard it introduces, which bit while updating the tests: **the type-argument count now selects the overload.** `arraySum[Int, Int]` picks arity one and then rejects three value arguments, where arity two wants `[Int, Int, Int]`. It is a compile error rather than a wrong query, and annotating the lambda parameters sidesteps it entirely, but it is worth knowing about.

## Arity, and why not by extending the old shape

I measured the obvious approach first. Extending the five parallel `Option` slots to five arities gives **911 lines, a seven-parameter base class, and four `Nothing`s in every arity-one construction** — and arity six would need the same surgery again.

So the lambda now lives in a `Lambda` ADT with one case per arity instead. That is strictly better on every axis:

- base class drops from **five type parameters to two**, and **eight of the eighteen nodes need none at all** (`ArrayAll` is one line where it was six)
- one more arity is a case in the ADT plus a rendering arm, not a change to eighteen nodes
- **861 lines** across both files, with the node section smaller than before
- still **no casts** — each rendering arm matches a concrete `Lambda` case, so the parameter types are recovered by the pattern rather than asserted

`Lambda` sits in `dsl` rather than beside the nodes in `dsl.column`. It renders nothing on its own, and the coverage gate over that package rightly expects everything there to — the gate caught this, and registering five meaningless entries would have been the wrong answer. It now sits with `Window` and `OrderingColumn`, which are likewise structures rather than functions.

## The docs page confirms two things this had been inferring

From `regular-functions/overview`: a lambda *"that accepts multiple arguments can also be passed"*, over *"several arrays of identical length"* — arity is open, and only the **lengths** must agree, not the types. And *"for some functions the first argument (the lambda function) can be omitted. In this case, identical mapping is assumed"* — which is exactly the eight families whose arity-one builder takes an `Option`. That split had been reverse-engineered from the existing signatures; it is now documented.

The same page mentions **`arrayFold`, which this DSL does not have at all.** Out of scope here — worth its own issue.

## Testing

487 unit tests green on 2.13.18 and 3.3.8, 186 integration tests against a live 25.3 server, including arity four and five executed end to end with results checked, and the wire forms for arity 4/5 verified directly against the server first.

## Updated breaking-change list

- `arrayXxx2` / `arrayXxx3` are gone; use `arrayXxx` at any arity.
- `HigherOrderFunction` goes from three type parameters to two; most nodes lose their parameters entirely.
- `arraySort` / `arrayReverseSort` result type changes from `Iterable[O]` to `Iterable[I1]` (a correctness fix — the server returns the first array reordered).
- `Lambda` is new public API in `com.crobox.clickhouse.dsl`.
- No `arraySplit` rename any more.

